PR #8, opened 2022-04-14
Comp 369 fix example in user guide

q-mat-beu:
- fix instructions in user-guide to visualize example circuit
- add note on requirements for building docs

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -47,7 +47,7 @@ Let's consider the following quantum circuit which prepares and measures a Bell 
     qc.cx(0, 1)
     qc.measure([0, 1], [0, 1])
 
-    qc.draw()
+    print(qc.draw(output='text'))
 
 ::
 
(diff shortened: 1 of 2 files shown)


docs/user_guide.rst
```diff
@@ -47,7 +47,7 @@ Let's consider the following quantum circuit which prepares and measures a Bell
     qc.cx(0, 1)
     qc.measure([0, 1], [0, 1])
 
-    qc.draw()
```
hay-k, line 50: There are other draw lines in the user guide that shall be modified like this.

docs/user_guide.rst
```diff
@@ -47,7 +47,7 @@ Let's consider the following quantum circuit which prepares and measures a Bell
...
+    print(qc.draw(output='text'))
```
hay-k, line 50: There are other draw lines in the user guide that shall be modified like this.

q-mat-beu: Done.

hay-k (approved):

PR #11, opened 2022-04-29
Enable user auth

ollityrkko:


setup.cfg
```diff
@@ -35,7 +35,7 @@ package_dir =
 install_requires =
     numpy
     qiskit ~= 0.34.1
-    iqm-client >= 1.10, < 2.0
+    iqm-client >= 2.0
```
hay-k, line 38:
```suggestion
    iqm-client >= 2.0, < 3
```

src/qiskit_iqm/iqm_provider.py
```diff
@@ -26,16 +26,30 @@ class IQMProvider:
...
         with open(settings_path, 'r', encoding='utf-8') as f:
-            self._client = IQMClient(url, json.loads(f.read()), username, api_key)
+            self._client = IQMClient(url, json.loads(f.read()), **user_auth_args)
 
     def get_backend(self) -> IQMBackend:
         """An IQMBackend instance associated with this provider.
         """
         return IQMBackend(self._client)
+
+    def close(self):
```
hay-k, line 52: Perhaps it is more convenient to have this method in `IQMBackend`, since `IQMProvider` is just used to generate backend objects and pass the client to them. We could do so that IQMProvider does not store client in the instance variable `self._client` at all. We could just save the arguments to ``__init__` as instance variables, and create the client object inside the `get_backend` method. Then `IQMProvider` will not need the close function at all, and the function could be moved to backend.

hay-k: Also, perhaps it could be renamed to `close_client` or something similar, since close is too general

setup.cfg
ollityrkko: fixed

hay-k (approved):

PR #18, opened 2022-06-22
Batch circuits.

q-mat-beu:
Experimental enabling of batch circuits in qiskit-on-iqm.

### DIFF
--- a/src/qiskit_iqm/iqm_backend.py
+++ b/src/qiskit_iqm/iqm_backend.py
@@ -15,7 +15,7 @@
 """
 from __future__ import annotations
 
-from typing import Union
+from typing import Optional, Union
 
 from iqm_client.iqm_client import IQMClient
 from qiskit import QuantumCircuit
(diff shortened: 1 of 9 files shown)


src/qiskit_iqm/iqm_backend.py
```diff
@@ -50,17 +50,27 @@ def max_circuits(self) -> int:
...
         shots = options.get('shots', self.options.shots)
 
-        circuit_serialized = serialize_circuit(circuit)
-        mapping_serialized = serialize_qubit_mapping(qubit_mapping, circuit)
+        circuits_serialized = [serialize_circuit(circuit) for circuit in circuits]
+        mappings_serialized = [
+            serialize_qubit_mapping(qubit_mapping, circuit)
+            for circuit in circuits
+        ]
+        if (
```
ollityrkko, line 66: What is the intended logic in this check?
Should it rather be ``any(mapping != mappings_serialized[0] ... ) or not mappings_serialized``?
When is ``mappings_serialized`` empty (or falsey)? It seems it always has the same length as ``circuits``.
Should it be `` or not mappings_serialized[0]`` instead?

q-mat-beu: Changed it and added a test for the length of `circuits` above.

PR #21, opened 2022-08-16
Support iqm-client 6.0

kukushechkin:


tests/test_iqm_backend.py
```diff
@@ -137,3 +137,11 @@ def test_error_on_empty_circuit_list(backend):
             qubit_mapping={},
             shots=42
         )
+
+
+def test_client_close(backend):
```
freetonik, line 142: Same nit-pickiness as in related `cirq-on-iqm` PR: maybe rename to `test_close_client` to be in line with the corresponding method name `IQMSampler.close_client()`.

freetonik (approved):

PR #27, opened 2022-09-26
Enable mypy support

freetonik:


src/qiskit_iqm/iqm_job.py
```diff
@@ -35,11 +36,11 @@ class IQMJob(JobV1):
         job_id: string representation of the UUID generated by IQM server
         **kwargs: arguments to be passed to the initializer of the parent class
     """
-    def __init__(self, backend: 'qiskit_iqm.IQMBackend', job_id: str, **kwargs):
+    def __init__(self, backend: 'qiskit_iqm.IQMBackend', job_id: str, **kwargs): # type: ignore
```
freetonik, line 39: Here we have to use `'qiskit_iqm.IQMBackend'` annotation due to circular dependencies, but mypy can't find that type.

src/qiskit_iqm/iqm_job.py
```diff
@@ -62,7 +63,7 @@ def _format_measurement_results(
         measurement_results: CircuitMeasurementResults,
         shape: tuple[int, int]
     ) -> list[str]:
-        formatted_results = {}
+        formatted_results:dict[int, np.ndarray] = {}
```
kukushechkin, line 66: Is this a linter suggestion to glue symbols around the colon?

freetonik: No, but it doesn't care one way or another. But I've updated it anyway, let's be consistent.

kukushechkin (approved):

PR #32, opened 2022-10-24
Transpiler target

hay-k:


docs/user_guide.rst
```diff
@@ -28,60 +28,72 @@ After installation Qiskit on IQM can be imported in your Python code as follows:
...
-    from qiskit.compiler import transpile
+    from qiskit import execute
+    from qiskit_iqm import IQMProvider
 
-    qc_decomposed = transpile(qc, basis_gates=['r', 'cz'])
+    provider = IQMProvider(iqm_server_url)
+    backend = provider.get_backend()
 
-    print(qc_decomposed.draw(output='text'))
+    job = execute(qc, backend, shots=1000)
```
hay-k, line 80: This currently usually does not work, and one has to use `optimization_level=0` to make it work. This shall be fixed by https://github.com/Qiskit/qiskit-terra/issues/8970

src/qiskit_iqm/iqm_job.py
```diff
@@ -82,7 +83,11 @@ def _format_measurement_results(
         ]
 
     def submit(self):
-        raise NotImplementedError('Instead, use IQMBackend.run to submit jobs.')
+        raise NotImplementedError(
+            'You shall never need to submit jobs by calling this method. When running circuits through IQMBackend, '
```
q-mat-beu, line 87:
```suggestion
            'You should never have to submit jobs by calling this method. When running circuits through IQMBackend, '
```

hay-k:
> Looks like the tests currently don't execute anything because they can't find the right environment. Tests are failing for me because the fixture `linear_architecture_3q` is missing.

Of course I forgot to commit the `conftest.py` 🤦‍♂️

q-mat-beu (approved):

PR #33, opened 2022-10-27
Methods for qubit name to index and back

hay-k:
- Makes `IQMBackend` more user friendly
- adds some important tests

### DIFF
--- a/src/qiskit_iqm/iqm_backend.py
+++ b/src/qiskit_iqm/iqm_backend.py
@@ -113,13 +113,15 @@ class IQMBackend(BackendV2):
             self.client.close_auth_session()
         self.client = None
 
-    def qubit_name_to_index(self, name: str) -> int:
-        """Given an IQM-style qubit name ('QB1', 'QB2', etc.) return the corresponding index in the register."""
-        return self._qb_to_idx[name]
+    def qubit_name_to_index(self, name: str) -> Optional[int]:
+        """Given an IQM-style qubit name ('QB1', 'QB2', etc.) return the corresponding index in the register. Returns
+        None is the given name does not belong to the backend."""
+        return self._qb_to_idx.get(name)
 
-    def index_to_qubit_name(self, index: int) -> str:
-        """Given an index in the backend register return the corresponding IQM-style qubit name ('QB1', 'QB2', etc.)."""
-        return self._idx_to_qb[index]
+    def index_to_qubit_name(self, index: int) -> Optional[str]:
+        """Given an index in the backend register return the corresponding IQM-style qubit name ('QB1', 'QB2', etc.).
+        Returns None if the given index does not correspond to any qubit in the backend."""
+        return self._idx_to_qb.get(index)
 
     def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
         """Serialize a quantum circuit into the IQM data transfer format.
--- a/src/qiskit_iqm/iqm_backend.py
+++ b/src/qiskit_iqm/iqm_backend.py
@@ -15,6 +15,7 @@
 """
 from __future__ import annotations
 
+import re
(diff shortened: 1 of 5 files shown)


tests/test_iqm_backend.py
```diff
@@ -61,6 +61,24 @@ def test_max_circuits(backend):
...
+def test_qubit_name_to_index_to_qubit_name(adonis_architecture_shuffled_names):
+    client = mock(IQMClient)
+    when(client).get_quantum_architecture().thenReturn(adonis_architecture_shuffled_names)
+    backend = IQMBackend(client)
+
+    assert backend.qubit_name_to_index('QB1') == 0
+    assert backend.qubit_name_to_index('QB2') == 1
+    assert backend.qubit_name_to_index('QB3') == 2
+    assert backend.qubit_name_to_index('QB4') == 3
+    assert backend.qubit_name_to_index('QB5') == 4
```
freetonik, line 73: To avoid the struggle when Apollo is needed :-)

```suggestion
    indices = list(range(0, 5))
    indices_from_qubits = list(backend.qubit_name_to_index(i) for i in indices)
    assert indices == indices_from_qubits
```

---

Alternatively, there is unittest's [assertSequenceEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertSequenceEqual), for which we can simplify:

```python
    indices = range(0, 5)
    indices_from_qubits = (backend.qubit_name_to_index(i) for i in indices)
    assertSequenceEqual(indices, indices_from_qubits)
```

freetonik, line 73: To avoid the struggle when Apollo is needed :-)

```suggestion
    indices = list(range(0, 5))
    qubit_names = ['QB1', 'QB2', 'QB3', 'QB4', 'QB5']
    indices_from_qubits = list(backend.qubit_name_to_index(q) for q in qubit_names)
    assert indices == indices_from_qubits
```

Alternatively, there is unittest's [assertSequenceEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertSequenceEqual).

hay-k: right, I was trying to be explicit, but I guess this is still pretty explicit as well.

src/qiskit_iqm/iqm_backend.py
```diff
@@ -106,6 +113,14 @@ def close_client(self):
             self.client.close_auth_session()
         self.client = None
 
+    def qubit_name_to_index(self, name: str) -> int:
+        """Given an IQM-style qubit name ('QB1', 'QB2', etc.) return the corresponding index in the register."""
+        return self._qb_to_idx[name]
```
freetonik, line 118: I understand that, by design, this should never fail, but I instinctively have a bad feeling about optimistically retrieving a value like this. I don't have good suggestions though :D

Perhaps there is a suitable exception it can throw in case key is not found?

hay-k: Do you mean scenarios like a user requesting an incorrect qubit name? I guess I can return `None` in that case

hay-k: Done

smite:
Can you add a section in the docs about how the qubit mapping / routing works? I mean from user-made circuits with arbitrary qregs, to routed circuits, to IQM qubit names.

hay-k:
> Can you add a section in the docs about how the qubit mapping / routing works? I mean from user-made circuits with arbitrary qregs, to routed circuits, to IQM qubit names.

I don't think it deserves a section, since most of it is typical functionality of Qiskit, and our documentation is not intended for teaching Qiskit. But I can perhaps add a sentence or two about IQM-style qubit names in the user guide and/or the docstring of `IQMBackend`

q-mat-beu (approved):

PR #38, opened 2023-01-18
Add an example to user guide for OpenQASM circuit execution.

q-mat-beu:


docs/user_guide.rst
```diff
@@ -167,6 +167,31 @@ We can also simulate the execution of the transpiled circuit before actually exe
 
     print(job.result().get_counts())
 
+In addition to circuits defined with Qiskit it is also possible to execute circuits defined
+in an OpenQASM 2.0 file:
```
hay-k, line 171:
```suggestion
If case you want to execute a circuit defined as an OpenQASM 2.0 file, you can load it into a Qiskit representation and then execute as normal:
```

PR #40, opened 2023-01-26
COMP-707 Update tox configuration

rbrazinskas:
- Supported Python interpreter versions are "pinned down" to 3.9 and 3.10
- Tox skips missing versions of Python interpreters (instead of raising an exception)
- Project is migrated to use <pyproject.toml> file as a single and only source of configuration (except of <AUTHORS.rst> file which is needed for <sphinx> document build)


pyproject.toml
```diff
@@ -1,24 +1,161 @@
...
+dependencies = [
+    "numpy",
+    "qiskit ~= 0.39.1",
+    "iqm-client >= 11.0, < 12.0"
+
+]
+
+[project.urls]
+homepage = "https://github.com/iqm-finland/qiskit-on-iqm"
+doumentation = "https://iqm-finland.github.io/qiskit-on-iqm"
```
ollityrkko, line 30: typo: documentation

ollityrkko (approved):
One typo, otherwise looks good

pyproject.toml
rbrazinskas: Good catch! Fixed.

ollityrkko (approved):

ollityrkko (approved):

PR #41, opened 2023-02-01
Add a simple example

olliahonen:
Adds a simple getting started example for the beginner's convenience

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -6,6 +6,28 @@ User guide
 This guide illustrates the main features of Qiskit on IQM. You are encouraged to run the demonstrated
 code snippets and check the output yourself.
 
+.. note::
+
+   At the moment IQM does not provide a quantum computing service open to the general public.
+   Please contact our `sales team <https://www.meetiqm.com/contact/>`_ to set up your access to an IQM quantum
+   computer.
+
+
+Hello, world!
+-------------
+
+Here's the quickest and easiest way to execute a small computation on an IQM quantum computer and to check that
+things are set up correctly:
+1. Download the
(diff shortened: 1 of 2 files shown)


docs/user_guide.rst
```diff
@@ -6,6 +6,26 @@ User guide
...
+
+Here's the quickest and easiest way to execute a small computation on an IQM quantum computer and check that
+things are set up correctly:
+
+1. Download the `bell_measure.py example file <https://github.com/iqm-finland/qiskit-on-iqm/tree/main/examples/bell_measure.py>`_
+2. Install Qiskit on IQM as instructed below (feel free to skip the import statement)
+3. Install Cortex CLI and log in as instructed in the `documentation <https://iqm-finland.github.io/cortex-cli/readme.html#installing-cortex-cli>`__
+4. Set the environment variable as instructed by Cortex CLI after logging in
+5. Run ``$ python bell_measure.py --server_url https://demo.qc.iqm.fi/cocos`` – replace the example URL with the correct one
+6. The output should show almost half of the measurements resulting in '00' and almost half in '11' – if this is the case, things are set up correctly!
```
q-mat-beu, line 27: Maybe clarify that this should be the case if one is connected against a real, calibrated QC.

olliahonen, line 27:
```suggestion
6. If you're connecting to a real quantum computer, the output should show almost half of the measurements resulting in '00' and almost half in '11' – if this is the case, things are set up correctly!
```

olliahonen: Done. I was sort of assuming connecting a to real QC is the default 🙂

PR #43, opened 2023-02-06
Build and publish docs for older versions

kukushechkin:


docs/conf.py
```diff
@@ -158,3 +169,13 @@
 
 # List of all bibliography files used.
 #bibtex_bibfiles = ['references.bib']
+
+# -- Options for sphinx_multiversion --------------------------------------------------
+# Only include certain tags (i.e. all tags except for ones listed below)
+# (technically a whitelist, but we treat it as blacklist by using negative lookahead regex `?!`)
+smv_tag_whitelist = r'^(?!(0\.[0-9]*)).*$'  # ignore all the versions before 4.0
```
q-mat-beu, line 176: Just out of curiosity, why?

q-mat-beu:
`test_settings.json` was probably committed by accident.

docs/conf.py
kukushechkin: Multiversion builds have a flaw — they do not respect requirements from old builds and only use requirements from the latest. And if older docs include some dependency no longer included in dev — build fails (`iqm_gates` package in this case). It can probably be fixed by pulling all the dependencies for every branch, but looks like too much to support older versions.

PR #47, opened 2023-03-15
Fix some broken docstrings.

smite:
* Fixes broken docstrings, updates others.
* More precise mapping of run statuses to Qiskit JobStatuses.


manzanillo (approved):

src/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -67,7 +71,7 @@ class IQMErrorProfile:
     two_qubit_gate_depolarizing_error_parameters: dict[str, dict[tuple[str, str], float]]
     single_qubit_gate_durations: dict[str, float]
     two_qubit_gate_durations: dict[str, float]
-    id_: Union[str, None] = None
+    id_: Optional[str] = None
```
a-auer, line 74: The parameter `id_` actually appears twice in the documentation. (Probably this was the case already before.)

hay-k: Looks like the documentation builder is confused by the underscore. Perhaps we can rename the variable to something else.

smite: Fixed.

q-mat-beu (approved):

PR #59, opened 2023-05-19
Add info about custom calibration set to user guide

freetonik:


CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 7.15
+============
+
+* Add info about custom calibration set to user guide `#58 <https://github.com/iqm-finland/qiskit-on-iqm/pull/59>`_
```
hay-k, line 8:
```suggestion
* Add info about custom calibration set to user guide `#59 <https://github.com/iqm-finland/qiskit-on-iqm/pull/59>`_
```

docs/user_guide.rst
```diff
@@ -258,6 +258,17 @@ connectivity and supported gates should match the 5-qubit Adonis architecture).
...
+----------------------
+
+Calibration set is a set of instrument and QPU settings which ensures the accuracy of computation. An IQM server uses
+the best available (at the time of execution) calibration set automatically. If you know an ID (UUID4) of a specific
+calibration set, you can provide it to Qiskit's ``execute`` method (which passes the value down to
+:meth:`.IQMBackend.run`):
+
+.. code-block:: python
+    job = execute(circuit, backend, shots=1000, calibration_set_id="f7d9642e-b0ca-4f2d-af2a-30195bd7a76d")
+
```
hay-k: I would move this into the `Running a quantum circuit on an IQM quantum computer` section, just under the sentence `Note that the code snippet above assumes that you have set the variable iqm_server_url.`, instead of making a separate subsection.

Also, I would phrase is something like this:

You can optionally set IQM backend specific settings as additional keyword arguments to the `execute` method (which passes the value down to
:meth:`.IQMBackend.run`). For example, an IQM server uses
the best available calibration set automatically, however if you know an ID (UUID4) of a specific calibration set that you want to use, you can provide it as follows:

freetonik: Good suggestion, thanks. Done (with minor edits).

hay-k (approved):

PR #61, opened 2023-05-22
Add ``circuit_duration_check``

freetonik:


jkotilahti (approved):

src/qiskit_iqm/iqm_provider.py
```diff
@@ -61,14 +61,19 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
 
         shots = options.get('shots', self.options.shots)
         calibration_set_id = options.get('calibration_set_id', self.options.calibration_set_id)
+        circuit_duration_check = options.get('circuit_duration_check', self.options.circuit_duration_check)
```
smite, line 64: Should we document all the `options` using a `Keyword Args:` section in the `run` docstring?

hay-k: The way Qiskit implements these options make it really hard to document anywhere. 
One can either set options by passing as keyword arguments to the `run` method, or one could set them on the backend instance via `backend.set_options`, and then call the `run` method without any options. Documenting them in the `run` method does not make them visible for the second use case. The ideal place to document these would be the `Options` object created in `_default_options` method, and users could check documentation from `backend.options`. Unfortunately `Options` class does not provide any way of specifying documentation/description of options.
Perhaps the best we can to is to override the `options` property of backend, without changing the implementation and add documentation there. I have been thinking to do this in my MR for adding the heralding option.

hay-k: Or perhaps it is better to document them in the IQMBackend class documentation itself

hay-k: OK, after trying to put it in multiple places (docstrings of `_default_options`, `options`, `run`, `IQMBackend`) and seeing how they look in the generated documentation, I came to the conclusion that the best place for them is the user guide. @freetonik if you merge your PR as is, I will add description of all options in user guide along with my PR (https://github.com/iqm-finland/qiskit-on-iqm/pull/63)

PR #65, opened 2023-06-02
Bugfix: Handle case where heralding returns zero shots

q-mat-beu:
This MR fixes two small issues:
* Handle case where heralding returns zero shots
* Handle cal set id both as str and UUID

### DIFF
--- a/src/qiskit_iqm/iqm_job.py
+++ b/src/qiskit_iqm/iqm_job.py
@@ -19,6 +19,7 @@ from collections import Counter
 from datetime import date
 from typing import Optional, Union
 import uuid
+import warnings
 
 from iqm_client import CircuitMeasurementResults, HeraldingMode, IQMClient, RunRequest, RunResult, Status
 import numpy as np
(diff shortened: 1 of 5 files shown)


src/qiskit_iqm/iqm_job.py
```diff
@@ -67,11 +68,15 @@ def _format_measurement_results(
         for k, v in measurement_results.items():
             mk = MeasurementKey.from_string(k)
             res = np.array(v, dtype=int)
-            # in Qiskit each measurement is a separate single-qubit instruction. qiskit-iqm assigns unique measurement
-            # key to each such instruction, so only one column is expected per measurement key.
-            if res.shape[1] != 1:
-                raise ValueError(f'Measurement result {mk} has the wrong shape {res.shape}, expected (*, 1)')
-            res = res[:, 0]
+            if len(v) == 0 and not expect_exact_shots:
+                warnings.warn('Received measurement results with zero results')
```
hay-k, line 72:
```suggestion
                warnings.warn(
                    'Received measurement results containing zero shots. '
                    'In case you are using non-default heralding mode, this could be because of bad calibration.'
                )
```

hay-k (approved):

PR #67, opened 2023-06-13
Fix a bug in the Hello world example

olliahonen:
The server URL argument was somehow named incorrectly.


freetonik:
This will need a separate MR to update the link to bell_measure.py in the docs.

PR #68, opened 2023-06-15
Update example script link

olliahonen:
Follow-up for #67


olliahonen:
Actually #67 could've been merged without releasing, and this would've been release 10.0.

freetonik (approved):

PR #69, opened 2023-06-20
Implement cancel method for IQMJob

jkotilahti:


jkotilahti:
(CI will fail until https://github.com/iqm-finland/iqm-client/pull/89 is merged)

src/qiskit_iqm/iqm_job.py
```diff
@@ -103,8 +111,19 @@ def submit(self):
...
-        raise NotImplementedError('Canceling jobs is currently not supported.')
+    def cancel(self) -> bool:
+        """Attempt to cancel the job.
+
+        Returns:
+            True if the job was cancelled successfully, False otherwise
+        """
+        try:
+            self._client.abort_job(uuid.UUID(self._job_id))
+            print(f'Job {self._job_id} cancelled successfully.')
```
q-mat-beu, line 122: We should not use `print`. In the success case I would print nothing and in the failure case I would use the `warning` module.

jkotilahti: Using `warnings.warn` instead of `print` now, also checking in unit tests that warning is produced when expected.

q-mat-beu (approved):

CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 11.0
```
q-mat-beu, line 5: I think here it is actually not breaking because we have used `JobStatus` all along, so I would suggest a minor version update.

jkotilahti: Updated version number.

q-mat-beu (approved):

PR #71, opened 2023-07-06
Support id gate for iqm backends.

### LINKED ISSUE #46: id gate cannot be expanded when compiled for target
If I use a circuit with an Identity gate the transpile rules don't know how to handle it as it says in the docs the gate "should not be optimized or unrolled (it is an opaque gate)". (https://qiskit.org/documentation/stubs/qiskit.circuit.library.IGate.html)

Minimal example to reporduce:
```
from qiskit import execute, QuantumCircuit
from qiskit_iqm import IQMFakeAdonis

circuit = QuantumCircuit(2)
circuit.i(q)
circuit.measure_all()

backend = IQMFakeAdonis()
job = execute(circuit, backend, shots=1000)
job.result().get_counts()
```
Output:
```
QiskitError: "Cannot unroll the circuit to the given basis, ['r', 'cz', 'measure']. Instruction id not found in equivalence library and no rule found to expand."
```

One solution would be to add identity gate to the basis.

q-mat-beu:
Closes #46

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -184,9 +184,10 @@ Let's examine its basis gates and the coupling map through the ``backend`` insta
 
 ::
 
-    Native operations of the backend: ['r', 'cz', 'measure']
+    Native operations of the backend: ['r', 'id', 'cz', 'measure']
     Coupling map of the backend: [[0, 2], [1, 2], [2, 3], [2, 4]]
 
+Note that for IQM backends the identiy gate 'id' is not actually a gate that is executed on the device and is simply omitted.
 At IQM we identify qubits by their names, e.g. 'QB1', 'QB2', etc. as demonstrated above. In Qiskit, qubits are
 identified by their indices in the quantum register, as you can see from the printed coupling map above. Most of the
 time you do not need to deal with IQM-style qubit names when using Qiskit, however when you need, the methods
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -184,10 +184,10 @@ Let's examine its basis gates and the coupling map through the ``backend`` insta
 
 ::
 
-    Native operations of the backend: ['r', 'id', 'cz', 'measure']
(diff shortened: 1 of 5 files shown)


hay-k (approved):

docs/user_guide.rst
```diff
@@ -184,9 +184,10 @@ Let's examine its basis gates and the coupling map through the ``backend`` insta
 
 ::
 
-    Native operations of the backend: ['r', 'cz', 'measure']
+    Native operations of the backend: ['r', 'id', 'cz', 'measure']
```
smite, line 187:
```suggestion
    Native operations of the backend: ['id', 'r', 'cz', 'measure']
```

docs/user_guide.rst
```diff
@@ -184,9 +184,10 @@ Let's examine its basis gates and the coupling map through the ``backend`` insta
...
     Coupling map of the backend: [[0, 2], [1, 2], [2, 3], [2, 4]]
 
+Note that for IQM backends the identiy gate 'id' is not actually a gate that is executed on the device and is simply omitted.
```
smite, line 190:
```suggestion
Note that for IQM backends the identiy gate ``id`` is not actually a gate that is executed on the device and is simply omitted.
```

tests/test_iqm_provider.py
```diff
@@ -195,8 +195,16 @@ def test_serialize_circuit_barrier(circuit, backend):
     assert circuit_ser.instructions[1].args == {}
 
 
+def test_serialize_circuit_id(circuit, backend):
+    circuit.r(theta=np.pi, phi=0, qubit=0)
+    circuit.id(0)
+    circuit_ser = backend.serialize_circuit(circuit)
+    assert len(circuit_ser.instructions) == 1
```
smite, line 202: Could also check that the one remaining instruction is the r :)

hay-k (approved):

PR #72, opened 2023-07-13
Enable option `memory` in IQMFakeBackend

freetonik:


src/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -283,9 +283,10 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
             raise ValueError("Empty list of circuits submitted for execution.")
 
         shots = options.get("shots", self.options.shots)
+        memory = options.get("memory", self.options.memory)
 
         # Create noisy simulator backend and run circuits
         sim_noise = AerSimulator(noise_model=self.noise_model)
-        job = sim_noise.run(circuits, shots=shots)
+        job = sim_noise.run(circuits, shots=shots, memory=memory)
```
q-mat-beu, line 290: Is there any reason not to pass all arguments in `options`?

PR #75, opened 2023-08-23
Fixed 2QB gate error depending on qubit order

a-auer:
Errors to two-qubit `CZ` gates were only applied for one specific order of the qubit arguments in the gate construction. But the error should be applied independent of this order since the gate is symmetric in control and target qubit.

Changes:

- Added reverse qubit order in error construction in `iqm_fake_backend.py`.
- Added assertion to test case `test_noise_model_contains_all_errors` in `test_iqm_fake_backend.py`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 10.8
+============
+
+* Fix two-qubit gate error construction in ``IQMFakeBackend``.
+
 Version 10.7
 ============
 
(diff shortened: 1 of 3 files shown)


q-mat-beu:
Looks good, but you also need a Changelog entry to trigger a release 🙂

a-auer:
> Looks good, but you also need a Changelog entry to trigger a release 🙂

Done.

q-mat-beu (approved):

PR #80, opened 2023-09-20
IQMBackend improvements

hay-k:
- Adds a new option called `circuit_callback` to `IQMBackend`. Description and motivation is in the user guide.
- Unknown options passed to backend now cause an error. Previously they were just silently ignored, which was not good. E.g. if you wanted to specify custom calset ID, but you made a typo in the name of the option, it would get ignored and the default calibration set would be still used. And you would be left wondering why your calibration set is not being used. Now you will immediately get error that you have not supplied a correct option.
- Handling of options is now improved a bit and the code is easier to read and consume.
- Fix the type of ``date`` in result object. Qiskit expects it to be string. It was datetime before.
- ``IQMBackend`` can now serialize `x`, `rx`, `y` and `ry` gates as well. More details and motivation in the docstring of the serialize function.
- Improved the layout of user guide.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -139,6 +139,17 @@ their current values using `backend.options`. Below table summarizes currently a
      - :py:class:`~iqm_client.iqm_client.HeraldingMode`
      - "zeros"
      - Heralding mode to use during execution. The default value is "none".
+   * - `circuit_callback`
+     - :py:class:`collections.abc.Callable`
+     - ``None``
+     - A function that accepts a list of :py:class:`qiskit.QuantumCircuit` instances and does not return anything.
+       When the backend receives circuits for execution, it will call this function (if provided) and pass those
+       circuits as argument. This may be useful in situations when you do not have explicit control over transpilation,
+       but need some information on how it was done. This can happen, for example, when you use pre-implemented
+       algorithms and experiments in Qiskit, where the implementation of the said algorithm or experiment takes care of
+       delivering correctly transpiled circuits to the backend. This callback method gives you a chance to look into
+       those transpiled circuits, extract any info you need.
+
 
 If the IQM server you are connecting to requires authentication, you will also have to use
 `Cortex CLI <https://github.com/iqm-finland/cortex-cli>`_ to retrieve and automatically refresh access tokens,
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -146,7 +146,8 @@ their current values using `backend.options`. Below table summarizes currently a
        but need some information on how it was done. This can happen, for example, when you use pre-implemented
        algorithms and experiments in Qiskit, where the implementation of the said algorithm or experiment takes care of
        delivering correctly transpiled circuits to the backend. This callback method gives you a chance to look into
-       those transpiled circuits, extract any info you need.
+       those transpiled circuits, extract any info you need. As a side effect, you can also use this callback to modify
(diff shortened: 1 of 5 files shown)


tests/test_iqm_provider.py
```diff
@@ -312,6 +354,31 @@ def test_run_with_heralding_mode_zeros(backend, circuit, submit_circuits_default
...
 
 
+def test_run_with_circuit_callback(backend, job_id, submit_circuits_default_kwargs):
+    qc1 = QuantumCircuit(3)
+    qc1.measure_all()
+    qc2 = QuantumCircuit(3)
+    qc2.r(np.pi, 0.3, 0)
+    qc2.measure_all()
+
+    def sample_callback(circuits) -> None:
```
q-mat-beu, line 364: You could add a check that the callback is actually called, e.g. by using some mock and asserting calls.

docs/user_guide.rst
```diff
@@ -112,33 +112,42 @@ and then call execution methods without specifying additional keyword arguments.
...
+
+       Enable or disable server-side circuit duration checks. The default value is `True`, which means if any job is
        estimated to take unreasonably long compared to the coherence times of the qubits, or too long in wall-clock
        time, the server will reject it. This option can be used to disable this behaviour. In normal use, the
        circuit duration check should always remain enabled.
    * - `heralding_mode`
-     - :py:class:`~iqm_client.iqm_client.HeraldingMode`
-     - "zeros"
-     - Heralding mode to use during execution. The default value is "none".
+     - Type: :py:class:`~iqm.iqm_client.iqm_client.HeraldingMode`, Example value: ``"zeros"``.
```
smite, line 137:
```suggestion
     - Type: :class:`~iqm.iqm_client.iqm_client.HeraldingMode`, Example value: ``"zeros"``.
```
The `py` domain is the default.

docs/user_guide.rst
```diff
@@ -112,33 +112,42 @@ and then call execution methods without specifying additional keyword arguments.
...
+
+       Heralding mode to use during execution. The default value is "none".
+   * - `circuit_callback`
+     - Type: :py:class:`collections.abc.Callable`, Example value: ``None``.
+
+       A function that accepts a list of :py:class:`qiskit.QuantumCircuit` instances and does not return anything.
```
smite, line 143:
```suggestion
     - Type: :class:`collections.abc.Callable`, Example value: ``None``.

       A function that accepts a list of :class:`qiskit.QuantumCircuit` instances and does not return anything.
```

smite: So the idea is that `circuit_callback` can modify the given circuits in place?
Should it instead return the circuits to be executed as `list[qiskit.QuantumCircuit]`, so it can either return what it was given as is, or edited copies?

hay-k: No, it is not intended for modifying circuits. Its main intention is to extract some info that user needs from circuits, i.e. only read the circuits. As a side effect they can also modify circuits in-place, but that's not the intended use for this option.

tests/test_iqm_provider.py
hay-k: Does it look good enough now?

q-mat-beu (approved):

docs/user_guide.rst
smite: OK, maybe just mention in the docstring that you're allowed to modify the circuits in place.

PR #88, opened 2023-12-18
Add name to backends.

q-mat-beu:


src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -97,14 +97,25 @@ class IQMFakeBackend(IQMBackendBase):
...
+        **kwargs,
+    ):
         super().__init__(architecture, **kwargs)
 
         self._validate_architecture_and_error_profile(architecture, error_profile)
         self.__architecture, self.__error_profile = architecture, error_profile
 
         self.noise_model = self._create_noise_model(architecture, error_profile)
 
+        if name:
```
smite, line 114:
```suggestion
        if name is not None:
```

q-mat-beu: Good point, changed!

rbrazinskas (approved):

PR #89, opened 2023-12-20
Bump dependencies and dev tool versions.

smite:
* ``iqm-client >= 15.2``

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,7 +40,7 @@ else:
 # -- General configuration ---------------------------------------------------
 
 # require a recent version of Sphinx
-needs_sphinx = '4.4'
+needs_sphinx = '6.2'
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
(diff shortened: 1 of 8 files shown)


docs/conf.py
```diff
@@ -40,7 +40,7 @@
 # -- General configuration ---------------------------------------------------
 
 # require a recent version of Sphinx
-needs_sphinx = '4.4'
+needs_sphinx = '6.2'
```
rbrazinskas, line 43: ``tox -e docs`` breaks on my local setup:

```
    title = caption % part
            ~~~~~~~~^~~~~~
TypeError: not all arguments converted during string formatting
```

I am able to fix it by changing ``extlinks`` dictionary in ``docs/conf.py`` to

```
extlinks = {
    'issue': ('https://github.com/iqm-finland/qiskit-on-iqm/issues/%s', 'issue %s'),
    'mr': ('https://github.com/iqm-finland/qiskit-on-iqm/pull/%s', 'MR %s'),
}
```

rbrazinskas (commented):
Small issue with buildin the docs (at least on my local Windows development environment. Otherwise LGTM

docs/conf.py
smite: Fixed.

rbrazinskas (approved):

tests/test_bell_measure_example.py
```diff
@@ -17,12 +17,13 @@
...
 
 import subprocess
+import sys
 
 
 def test_bell_measure_example_call():
     """Test that example script runs and fails at establishing a connection."""
     with subprocess.Popen(
-        ("python", "examples/bell_measure.py", "--cortex_server_url", "https://not.a.real.domain"),
+        (sys.executable, "examples/bell_measure.py", "--cortex_server_url", "https://not.a.real.domain"),
```
rbrazinskas, line 26: @kukushechkin, I would be really curious to understand why did the test break on Windows. Any ideas? :-)

kukushechkin: I guess the the "parent" python environment (comes from tox I guess?) is different from whatever we call by just calling "python" (comes from venv I guess?). This was it seems to be more consistent, it is not any random python, it is the same python that runs the tests.

kukushechkin (approved):

PR #90, opened 2024-01-22
Add support for the move gate and new the architecture model 

thaapasa:


src/iqm/qiskit_iqm/fake_backends/fake_adonis.py
```diff
@@ -20,30 +20,35 @@
...
+        qubit_connectivity=[['QB1', 'QB3'], ['QB2', 'QB3'], ['QB3', 'QB4'], ['QB3', 'QB5']],
     )
     error_profile = IQMErrorProfile(
-        t1s={"QB1": 27000.0, "QB2": 33000.0, "QB3": 25000.0, "QB4": 40000.0, "QB5": 25000.0},
-        t2s={"QB1": 20000.0, "QB2": 26000.0, "QB3": 23000.0, "QB4": 26000.0, "QB5": 7000.0},
+        t1s={'QB1': 27000.0, 'QB2': 33000.0, 'QB3': 25000.0, 'QB4': 40000.0, 'QB5': 25000.0},
+        t2s={'QB1': 20000.0, 'QB2': 26000.0, 'QB3': 23000.0, 'QB4': 26000.0, 'QB5': 7000.0},
         single_qubit_gate_depolarizing_error_parameters={
-            "phased_rx": {"QB1": 0.0006, "QB2": 0.0054, "QB3": 0.0001, "QB4": 0.0, "QB5": 0.0005}
+            'prx': {'QB1': 0.0006, 'QB2': 0.0054, 'QB3': 0.0001, 'QB4': 0.0, 'QB5': 0.0005}
```
thaapasa, line 37: I noticed now that currently the code base has used both single and double quotes for strings. Linter requires that the same quotes are used within a file, but not across the entire project. I changed these because I had to fix this file and the other files I worked with had used single quotes.

src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -44,28 +46,36 @@ def get_num_or_zero(name: str) -> int:
...
-                for pair in architecture.qubit_connectivity
-                for qb1, qb2 in (pair, pair[::-1])
-            },
-        )
-        target.add_instruction(Measure(), {(qb_to_idx[qb],): None for qb in architecture.qubits})
+        if 'cz' in operations:
+            target.add_instruction(
+                CZGate(),
+                {
+                    (qb_to_idx[qb1], qb_to_idx[qb2]): None
```
kukushechkin, line 68: This is implicitly duplicated knowledge which operation is sensitive to the order of qubits and which is not, `is_directed_operation` should also be used.

src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -92,10 +102,30 @@ def validate_compatible_architecture(self, architecture: QuantumArchitectureSpec
...
+        ops_match = compare_operations(architecture.operations, self._quantum_architecture.operations)
 
         self_connectivity = set(map(frozenset, self._quantum_architecture.qubit_connectivity))  # type: ignore
         target_connectivity = set(map(frozenset, architecture.qubit_connectivity))  # type: ignore
         connectivity_match = self_connectivity == target_connectivity
 
         return qubits_match and ops_match and connectivity_match
+
+
+def compare_operations(ops1: dict[str, list[list[str]]], ops2: dict[str, list[list[str]]]) -> bool:
```
kukushechkin, line 114: I would add unit test for this.

pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
     "numpy",
     "qiskit >= 0.45.1, < 0.46",
     "qiskit-aer >= 0.13.1, < 0.14",
-    "iqm-client >= 15.2, < 16.0"
+    "iqm-client >= 16.0a1, < 17.0"
```
freetonik, line 25: It's possible that while the custom branch is still not merged into master, 2 or more new stable major versions of IQM Client are released (e.g. 16.0 and then 16.1); this range will allow qiskit-on-iqm to fail. Perhaps, it should be a strict dependency on `16.0a1` here, or put the range for `a`-packages only (e.g. `>= 16.0a1, <16.99a1`, but I'm not sure that works)

src/iqm/qiskit_iqm/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    r.sort()
+    return r
+
+
+def is_multi_qubit_operation(operation: str):
+    """Checks whether the given operation is a multi-qubit operation"""
+    return operation in ['cz', 'move']
+
+
+def is_directed_operation(operation: str):
```
kukushechkin, line 33: Would it be better to make this an attribute of `Gate`, maintain only mapping between string names and classes and check in a generic way if qubits order is important?

tests/move_architecture/test_architecture.py
```diff
@@ -0,0 +1,107 @@
...
+        'cz:2',
+        'measure:1',
+    ]
+    check_instruction(backend.instructions, 'r', [(0,), (1,), (2,), (3,), (4,)])
+    check_instruction(backend.instructions, 'measure', [(0,), (1,), (2,), (3,), (4,)])
+    check_instruction(backend.instructions, 'id', [(0,), (1,), (2,), (3,), (4,)])
+    check_instruction(backend.instructions, 'cz', [(0, 2), (2, 0), (1, 2), (2, 1), (3, 2), (2, 3), (4, 2), (2, 4)])
+
+
+def check_instruction(
```
kukushechkin, line 100: This can be part of iqm-client validation logic.

src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -92,10 +102,30 @@ def validate_compatible_architecture(self, architecture: QuantumArchitectureSpec
...
+    """Compares the given operation sets defined by the quantum architecture. Returns true if they are the same."""
```
freetonik, line 115: Could you clarify `the same` in the docstring to be more precise, e.g. either "equivalent" or "identical".

tests/move_architecture/test_move_circuit.py
```diff
@@ -0,0 +1,74 @@
...
+        'move:6,0',
+        # cz(0, 3)
+        'cz:0,3',
+        # cz(4, 0) is optimized before h(6)
+        'cz:4,0',
+        # h(6)
+        'prx:6',
+        'prx:6',
+        # move (6, 0)
+        'move:6,0',
```
kukushechkin, line 70: Is it actually so that this second 6->0 move gate will be treated in the other direction?

tests/move_architecture/test_architecture.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def test_parse_move_architecture():
+    """Tests that the quantum architecture model can be parsed"""
+    arch = QuantumArchitecture(**move_architecture_json)
+    check_arch(arch.quantum_architecture)
+
+
+def check_arch(arch: QuantumArchitectureSpecification):
+    """Verifies architecture is as expected"""
```
freetonik, line 38:
```suggestion
    """Verifies architecture is as expected."""
```

freetonik: And same for other new docstrings, just to be consistent with existing docstrings in the codebase.

tests/move_architecture/test_move_circuit.py
kukushechkin: Ah, right, the **generated pulse** is the same, but the direction of move depends on where is the ground state.

thaapasa: Based on discussions with William and Nicole, the move operation is symmetric, so it switches state both ways. It is the convention to define this operation with the qubit first, and resonator second. That's why it's looks the same although the effect can be considered as going into one direction or the other.

src/iqm/qiskit_iqm/utils.py
thaapasa: I'll need this operation in the IQM client side as well, so now I'm considering extending the SUPPORTED_INSTRUCTIONS dict and adding the info there, then we could read that information in this adapter as well. That dict is not yet read here so let's see if we need to add a more explicit way of querying for that info as opposed to just reading the dict directly.

pyproject.toml
thaapasa: Changed this to be explicit

tests/move_architecture/test_architecture.py
thaapasa: Fixed documentation

src/iqm/qiskit_iqm/iqm_backend.py
thaapasa: Sure, fixed that

tests/move_architecture/test_architecture.py
thaapasa: I've now added validation logic to IQM client that checks that operations are targeting correct qubits (loci are defined in the architecture specification), but this check is for testing that the qiskit backend is configured correctly, so that qiskit's algorithms work properly. The Instruction here is qiskit's Instruction. So I need to have these checks here as well.

src/iqm/qiskit_iqm/iqm_backend.py
thaapasa: I've now generalized qiskit target configuration (how to define connectivity) and this info is now queried from IQM client.

pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
     "numpy",
     "qiskit >= 0.45.1, < 0.46",
     "qiskit-aer >= 0.13.1, < 0.14",
-    "iqm-client >= 16.0, < 17.0"
+    "iqm-client >= 17.0, < 18.0"
```
freetonik, line 25:
```suggestion
    "iqm-client >= 17.1, < 18.0"
```

To avoid the possibility of using the version of IQM Client which had quantum architecture a compatibility bug.

kukushechkin (approved):

PR #93, opened 2024-02-26
Fix Git tag selection regex in sphinx configuration

rbrazinskas:
Use regex that selects Git tags of the format MAJOR.MINOR that are >= 4.0. Also filter out all tags that do not strictly match format INTEGER.INTEGER. For example, "12.a1" would be filtered out

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 12.1
+============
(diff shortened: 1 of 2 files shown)


rbrazinskas:
Closing this PR, because we just have dropped the support for multiversion documentation builds; and the bug was in sphinx-multiversion-contrib configuration

PR #95, opened 2024-03-14
Clarify the purpose of max_circuits property

rbrazinskas:


src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -66,16 +66,22 @@ def _default_options(cls) -> Options:
 
     @property
     def max_circuits(self) -> Optional[int]:
+        """Set :attr:`max_circuits` property to a custom value.
```
smite, line 69: This is now the getter, not the setter

smite:
```suggestion
        """Maximum number of circuits that should be run in a single batch by ``qiskit_experiments``.
```

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -66,16 +66,22 @@ def _default_options(cls) -> Options:
...
+
+        Generally, this property is set automatically to `None` during the construction of backend instance. `None`
+        means there is no limit to the maximum number of circuits that can be run in a single batch.
```
smite, line 72: Can, or should? See the last sentence below.

rbrazinskas: Could? :-)

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -66,16 +66,21 @@ def _default_options(cls) -> Options:
...
+        """Maximum number of circuits that should be run in a single batch by ``qiskit_experiments``.
+
+        Generally, this property is set automatically to `None` during the construction of backend instance. `None`
+        means there is no limit to the maximum number of circuits that can be run in a single batch.
+
+        However, in certain situations, user wants to have control over how many circuits are included in a single
+        batch. By setting this property user can easily override the default value with a custom one.
+
+        Besides the above use case, this property is not used to limit the amount of circuits that is being sent to
+        the backend.
```
smite, line 78:
```suggestion
        """Maximum number of circuits that should be run in a single batch.
        
        Currently there is no hard limit on the number of circuits that can be executed in a single batch/job.
        However, some libraries like Qiskit Experiments use this property to split multi-circuit computational
        tasks into multiple baches/jobs.

        The default value is ``None``, meaning there is no limit. You can set it to a specific integer
        value to force these libraries to execute at most that many circuits in a single batch.
```

smite (changes requested):

ollityrkko (approved):

rbrazinskas (commented):
LGTM

smite (approved):

PR #97, opened 2024-03-20
Update user guide examples

ollityrkko:


docs/user_guide.rst
rbrazinskas, line 1: Just an idea, perhaps we could also unify the way we create a `provider`. Instead of having two different ways:

`provider = IQMProvider(iqm_server_url)`

and

`provider = IQMProvider("https://demo.qc.iqm.fi/cocos/")`

should we instead have

'iqm_server_url = "https://demo.qc.iqm.fi/cocos/"

provider = IQMProvider(iqm_server_url)'

everywhere?

rbrazinskas (approved):
One optional suggestion, otherwise LGTM

PR #99, opened 2024-04-10
Job monitoring bug fix

Aerylia:


src/iqm/qiskit_iqm/iqm_job.py
```diff
@@ -184,6 +184,21 @@ def status(self) -> JobStatus:
...
+            return ``None``. The ``refresh`` argument is ignored.
+
+        Args:
+            refresh: If ``True``, re-query the server to get the latest value.
+                Otherwise return the cached value.
+
+        Returns:
+            Position in the queue or ``None`` if position is unknown or not applicable.
+        """
+
```
kukushechkin, line 201: I would add an explicit `return None` and a comment that this is actually not implemented and not just an accident.

ollityrkko (approved):

src/iqm/qiskit_iqm/iqm_job.py
kukushechkin: Where does it complain about the unused argument? If it is truly unused in the place where `queue_position` is called, it should either be `_ = queue_position()` or explicitly disabled with `#pylint: disable` there.

jkotilahti:
Just adding a note here to not merge this until https://github.com/iqm-finland/qiskit-on-iqm/pull/101 has been merged, because otherwise publishing this version to PyPI will not work.

tests/test_iqm_job.py
```diff
@@ -255,3 +257,20 @@ def test_result_with_timestamps(job, iqm_result_two_registers, iqm_metadata_with
     assert job.metadata.pop('timestamps') == iqm_metadata_with_timestamps.get('timestamps')
     assert 'timestamps' in result._metadata
     assert result.timestamps == iqm_metadata_with_timestamps.get('timestamps')
+
+
+def test_job_monitor(job, iqm_metadata):
+    client_result = RunResult(status=Status.READY, metadata=iqm_metadata)
+    run_responses = [
```
kukushechkin, line 264: There is no assertions in the test, please, add what is actually tested, e.g . just "job_monitor should not raise" or we actually want to verify contents of `monitor_string`? As far as I can see, it would be enough to test handling of every possible `Status ` and not to care about the monitoring string:

```suggestion
    run_responses = [RunStatus(status=s.value) for s in Status]
```

kukushechkin (approved):

PR #101, opened 2024-04-12
Fix publishing to PyPI

jkotilahti:
Sphinx roles like :file: are not allowed in README.rst which becomes the long_description in PyPI.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,7 +41,7 @@ The documentation of the latest Qiskit on IQM release is available
 Jump to our `User guide <https://iqm-finland.github.io/qiskit-on-iqm/user_guide.html>`_
 for a quick introduction on how to use Qiskit on IQM.
 
-You can build documentation for any older version locally by cloning the Git repository, checking out the 
+You can build documentation for any older version locally by cloning the Git repository, checking out the
 corresponding tag, and running the docs builder. For example, to build the documentation for version ``12.2``:
 
 .. code-block:: bash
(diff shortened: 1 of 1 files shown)


ollityrkko:
Should the CHANGELOG be updated as well?

jkotilahti:
@ollityrkko my idea was to try merging this without changelog update and try to use it to publish 13.3 to PyPI. But if it doesn't work, then there should be a new MR just updating changelog to 13.4. I think updating the changelog in this MR would mean that 13.3 cannot be published to PyPI.  Not sure what is the best way to do this.

jkotilahti:
Actually, I guess publishing 13.3 to PyPI with this change is not possible anyway since the publishing job is tied to the old commit. So better to just update the changelog here as you suggested. But this will mean that 13.3 will not be published to PyPI.

I updated the changelog, and rebased on `main` which I forgot to do previously.

rbrazinskas:
> Actually, I guess publishing 13.3 to PyPI with this change is not possible anyway since the publishing job is tied to the old commit. So better to just update the changelog here as you suggested. But this will mean that 13.3 will not be published to PyPI.
> 
> I updated the changelog, and rebased on `main` which I forgot to do previously.

@jkotilahti If skipping version 13.3 on PyPI is an issue, I guess we could try to build and upload it manually. Let me know if you need help with that

rbrazinskas (approved):
LGTM. PyPI is able now to render README as show it as "Project descriprion"

PR #109, opened 2024-07-26
Add fake backend for IQM Deneb.

Jami-ronkko:


manzanillo (commented):
From a noise model perspective this definitely serves the purpose. However, I want to raise the question if it should behave similiarly in how Deneb behaves when the resonator is populated.
Probably the proposed solution also doesn't quite do the trick as there is only the requirement of the resonator to not be populated. My proposal just does checking if there was a MOVE operation performed before.

src/iqm/qiskit_iqm/fake_backends/fake_deneb.py
```diff
@@ -0,0 +1,130 @@
...
+"""
+from iqm.iqm_client import QuantumArchitectureSpecification
+from iqm.qiskit_iqm.fake_backends.iqm_fake_backend import IQMErrorProfile, IQMFakeBackend
+
+
+def IQMFakeDeneb() -> IQMFakeBackend:
+    """Return IQMFakeBackend instance representing IQM's Deneb architecture.
+    Note: swaps are used in place of the move gates to enable the (noisy) simulation with AerSimulator.
+    """
+    MOVE = "swap"
```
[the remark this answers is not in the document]
Jami-ronkko: Yes, good idea. It is a bit awkward that we need to include the "swap" and "move" in the operations of IQMFakeDeneb (former for the noise model and latter as native gates of transpiler).

Unless someone figures better way, I'll keep it and add a more detailed comment.

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -292,11 +303,26 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
...
+        class moves_to_swaps(TransformationPass):
+            """Change moves to swaps. It is assumed that moves are correctly placed."""
+
+            def run(self, dag):
+                for node in dag.op_nodes():
+                    if node.op.name == "move":
+                        move_as_swap = QuantumCircuit(2)
+                        move_as_swap.swap(0, 1)
+                        dag.substitute_node_with_dag(node, circuit_to_dag(move_as_swap))
+                return dag
```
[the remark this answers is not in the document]
Jami-ronkko: Good idea!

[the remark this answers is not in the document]
manzanillo: Okay, but then let's rely on the existing transpiler steps, one mom

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -292,11 +303,26 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
...
             ValueError: If empty list of circuits is provided.
         """
-        circuits = [run_input] if isinstance(run_input, QuantumCircuit) else run_input
+        circuits_aux = [run_input] if isinstance(run_input, (QuantumCircuit, IQMCircuit)) else run_input
 
-        if len(circuits) == 0:
+        if len(circuits_aux) == 0:
             raise ValueError("Empty list of circuits submitted for execution.")
 
+        class moves_to_swaps(TransformationPass):
```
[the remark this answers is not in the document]
Aerylia: The above code example is not valid for Deneb. Execute calls "transpile" followed by "run", but "transpile" is not defined for Deneb because of the Move. You need to use "transpile_to_iqm" followed by run to have the correct checks and balances.

src/iqm/qiskit_iqm/fake_backends/fake_deneb.py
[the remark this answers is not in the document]
Aerylia: What is the reason to define the MOVE as swap and not as the MoveGate that has been implemented with a mathematical definition (being a swap matrix)? The MoveGate should implement any missing features to allow usage in the AerSimulator so that it runs the same code and the transpile pass Move to Swaps isn't needed.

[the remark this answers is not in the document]
Jami-ronkko: Initially, I tried to do as you suggest but didn't succeed. I tried chancing the _define() of MoveGate to swap instead of the three cnots that it is currently. Then I used only 'move' in the noise model, but got error that 'move' is not recognized by AerSimulator.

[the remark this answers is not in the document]
Aerylia: I looked into this a bit further. The Aer simulator does not use the Gate definition to determine the unitary. I recommend to transpile MOVE gates to unitary gates where the unitary is a static field of the MoveGate class. And transpile to Move-to-Unitary following https://qiskit.github.io/qiskit-aer/tutorials/4_custom_gate_noise.html

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -292,11 +303,53 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
...
+                        else: 
+                            # MOVE OUT was performed
+                            qubits_involved_in_last_move = None
+                            
+                        move_as_swap = QuantumCircuit(2)
+                        move_as_swap.swap(0, 1)
+                        dag.substitute_node_with_dag(node, circuit_to_dag(move_as_swap))
+
+                    if qubits_involved_in_last_move is not None:
+                        raise ValueError("The following resonators are still holding qubit states " +
```
Aerylia, line 341: This if-statement is on the wrong level. Should be after the loop.

tests/fake_backends/test_fake_deneb.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_iqm_fake_deneb_connectivity(deneb_coupling_map):
+    backend = IQMFakeDeneb()
+    assert set(backend.coupling_map.get_edges()) == deneb_coupling_map
+
+
+def test_iqm_fake_deneb_noise_model_instantiated():
+    backend = IQMFakeDeneb()
+    assert isinstance(backend.noise_model, NoiseModel)
```
Aerylia, line 35: These tests don't test any functionality. The bug in the run transpile pass was not caught by it. I recommend some more use-case specific tests, like checking a GHZ circuit simulation and checking whether incorrect usage is caught.

Aerylia (changes requested):
Changes needed:
- Don't rely on "Swap" but get all information from the MoveGate class so that it can be easily adjusted. 
- Transpile move gates to Unitary gates so that we can later define a more precise unitary for the MoveGate
- More extensive user driven testing.

src/iqm/qiskit_iqm/fake_backends/fake_deneb.py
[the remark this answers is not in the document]
Aerylia: Yeah, I'm not sure if we just messaged past each other, but I think the better implementation is to have a static unitary field in MoveGate (nothing to do with the normal Qiskit API) and use that to transpile the MoveGates in the DAG to unitary gates that use the static field. That way, if we adjust the MoveGate description, the simulator can automatically pick it up and we don't need to remember to adjust it there. 

Given `noise_model = NoiseModel(basis_gates=["r", "cz", "unitary"])`
For example:
```
move_as_unitary = QuantumCircuit(2)
move_as_unitary.unitary(MoveGate.unitary,0, 1)
dag.substitute_node_with_dag(node, circuit_to_dag(move_as_unitary))
```

Alternatively you could do something like (not sure about compositionality of these functions)
```
unitary = Operator(dag_to_circuit(DAGCircuit)).to_matrix() 
```
Or
```
unitary = MoveGate(node.args).to_matrix()
```

That way, the simulator is not relying on the SWAP gate implementation specifically, but uses the specifics of the MoveGate class.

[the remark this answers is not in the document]
Jami-ronkko: Great, I was just going to ask exactly this. Do you prefer me adding static unitary to MoveGate or using one of the bottom cells to make the unitary in the IQMFakeBackend class?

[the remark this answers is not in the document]
Jami-ronkko: I pushed commit where I added unitary to MoveGate and use it for the noisy implementation.

tests/fake_backends/test_fake_deneb.py
Jami-ronkko: I added that Stefan did yesterday, which test the move logic. I also did two tests, checking noisy execution and that transpile_to_IQM works with the IQMFakeDeneb.

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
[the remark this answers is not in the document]
manzanillo: It works perfectly fine for a valid cricuit, however it would  check my code during compilation on server side.

```
    no_qubits = 6 
    comp_r = QuantumRegister(1, 'comp_r') # Computational resonator
    q = QuantumRegister(no_qubits, 'q') # Qubits 
    c = ClassicalRegister(no_qubits, 'c') # Classical register, used for readout
    qc = IQMCircuit(comp_r, q, c)

    qc.h(1)
    qc.x(2)
    qc.move(1,0)
    qc.cz(2,0)
    qc.move(1,0)
    qc.measure(q,c)

    job = execute(qc, backend, shots=1000)

    res=job.result()
    counts=res.get_counts()
    print(counts)
```

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -292,15 +303,60 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
...
+        if len(circuits_aux) == 0:
             raise ValueError("Empty list of circuits submitted for execution.")
 
+        class moves_to_swaps(TransformationPass):
+            """Checks for validity of move gates and changes them to swaps."""
+
+            def run(self, dag):
+                move_as_unitary = IQMCircuit(2)
+                move_as_unitary.unitary(MoveGate().unitary, [0, 1], label = "move")
+                move_as_unitary_dag = circuit_to_dag(move_as_unitary)
```
manzanillo, line 317: we can remove this part and rely on transpile now that we have a unitary defined
```suggestion
```

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -292,15 +303,60 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
...
+                        elif qubits_involved_in_last_move != node.qargs:
+                            raise ValueError("Cannot apply MOVE on 'QB"+ str(node.qargs[0].index + 1) + 
+                                             "' because COMP_R already holds the state of 'QB" +
+                                             str(qubits_involved_in_last_move[0].index + 1) + 
+                                             "'.")
+                        else: 
+                            # MOVE OUT was performed
+                            qubits_involved_in_last_move = None
+                            
+                        dag.substitute_node_with_dag(node, move_as_unitary_dag)
```
manzanillo, line 340: this we also get rid off then
```suggestion
```

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -292,15 +303,60 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
...
+
+
+        circuits = []
+        for circ in circuits_aux:
+            circuits.append(moves_to_swaps()(circ))
         shots = options.get("shots", self.options.shots)
 
         # Create noisy simulator backend and run circuits
         sim_noise = AerSimulator(noise_model=self.noise_model)
         job = sim_noise.run(circuits, shots=shots)
```
manzanillo, line 358:
```suggestion
        transpiled_circuits = transpile(circuits, sim_noise)
        job = sim_noise.run(transpiled_circuits, shots=shots)
```

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -22,13 +22,17 @@
 from typing import Any, Optional, Union
 
 from qiskit import QuantumCircuit
```
manzanillo, line 24:
```suggestion
from qiskit import QuantumCircuit, transpile
```

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -22,13 +22,17 @@
...
+from qiskit.converters import circuit_to_dag
```
manzanillo, line 25: Unneeded for this solution
```suggestion
```

manzanillo (approved):
Looks good from my side, I would just rely on the transpile feature from now as the unitary is already defined for the MOVE gate rather than exchanging the MOVE gates in the fake backend

Jami-ronkko:
I commited Stefan's suggestions to use transpile to implement the moves as unitaries (now that the unitary is defined in MoveGate). Noise model is also able to add noise to the moves without needing to use the label.

src/iqm/qiskit_iqm/move_gate.py
```diff
@@ -60,3 +65,5 @@ def _define(self):
             qc._append(instr, qargs, cargs)
```
manzanillo, line 65: I propose to change these lines 63-65 by removing them and defining the gate directly with its unitary:

```suggestion
        qc.unitary(self.unitary, [q[0], q[1]], label="move")
```

Aerylia: I agree.

Aerylia (changes requested):
With the proposed change from Stefan, this looks good.

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -221,8 +223,13 @@ def _create_noise_model(
         """
         Builds a noise model from the attributes.
         """
+
         noise_model = NoiseModel(basis_gates=["r", "cz"])
 
+        if architecture.name == "Deneb":
+            # noise_model.add_basis_gates(["move"]) does not work because it checks against a list of standard gates
+            noise_model = NoiseModel(basis_gates=["r", "cz", "move"])
+
```
Aerylia, line 232: Get the basis gates from the architecture, so it also works for future devices.
```suggestion
        iqm_to_qiskit_gates = {iqm: qiskit for iqm, qiskit in IQM_TO_QISKIT_GATE_NAME}
        for iqm_gate in architecture.operations.keys():
            if iqm_gate not in ["measure", "barrier"] + iqm_to_qiskit_gates.keys():
                iqm_to_qiskit_gates[iqm_gate] = iqm_gate
        noise_model = NoiseModel(basis_gates=iqm_to_qiskit_gates.values())
```

Aerylia: Which would require the use of IQM_TO_QISKIT_GATE_NAME below to use iqm_to_qiskit_gates instead.

src/iqm/qiskit_iqm/move_gate.py
```diff
@@ -55,8 +59,7 @@ def _define(self):
 
         q = QuantumRegister(2, "q")
         qc = QuantumCircuit(q, name=self.name)
-        rules = [(CXGate(), [q[0], q[1]], []), (CXGate(), [q[1], q[0]], []), (CXGate(), [q[0], q[1]], [])]
-        for instr, qargs, cargs in rules:
-            qc._append(instr, qargs, cargs)
+
+        qc.unitary(self.unitary, [q[0], q[1]], label="move")
```
Aerylia, line 63:
```suggestion
        qc.unitary(self.unitary, [q[0], q[1]], label=self.name)
```

src/iqm/qiskit_iqm/move_gate.py
```diff
@@ -55,8 +59,7 @@ def _define(self):
 
         q = QuantumRegister(2, "q")
         qc = QuantumCircuit(q, name=self.name)
```
Aerylia, line 61: Attempt to keep track of a potential given label.
```suggestion
        qc = QuantumCircuit(q, name=self.label if self.label else self.name)
```

Aerylia (approved):

Issue #108, opened 2024-07-23
Jobs timing out


ahojukka5:
I get APITimeoutError: The job didn't finish in 900 seconds.

By looking here (https://github.com/iqm-finland/iqm-client/blob/main/src/iqm/iqm_client/iqm_client.py#L433), `timeout_secs` is defined to be `DEFAULT_TIMEOUT_SECONDS`, but there's no way to control that argument from `qiskit_iqm`: (https://github.com/iqm-finland/qiskit-on-iqm/blob/main/src/iqm/qiskit_iqm/iqm_job.py#L130). 

APITimeoutError raises when the number of circuits and shots is big enough (64 circuits with 16384 shots is enough).

Aerylia:
Thank you for pointing out this feature discrepancy between iqm_client and qiskit_iqm's `IQMJob`. It has been added to our backlog (SW-381).

For now, the workaround is to use the `IQMClient` class directly and using `submit_circuits()`. This gives a `job_id` that you can use to construct the `IQMJob` using the `backend` and this `job_id`. You can make the Qiskit QuantumCircuit object into an iqm_client Circuit object using `backend.serialize_circuit(circuit)`. You might need to pass a `qubit_mapping` dict to make sure that the qubit names in the circuit fit those of the device. For example:

```
client = IQMClient(url="your url", **some_user_auth_kwargs)
job_id = client.submit_circuits([backend.serialize_circuit(circuit)], qubit_mapping={'1': "QB1", '2': "QB2"})
job = IQMJob(backend, job_id)
```

PR #112, opened 2024-08-02
COMP-1468 and SW-381

Aerylia:
Added the integration of the new feature of iqm-client: Compiler option MOVE gate validation bypassing (https://github.com/iqm-finland/iqm-client/pull/124) (COMP-1468)

Added support for passing Job timeout to iqm-client: #108 (SW-381)

### DIFF
--- a/src/iqm/qiskit_iqm/iqm_provider.py
+++ b/src/iqm/qiskit_iqm/iqm_provider.py
@@ -24,7 +24,7 @@ import numpy as np
 from qiskit import QuantumCircuit
 from qiskit.providers import JobStatus, JobV1, Options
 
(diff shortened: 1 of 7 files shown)


src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -91,6 +91,16 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
             raise ValueError('Empty list of circuits submitted for execution.')
 
         unknown_options = set(options.keys()) - set(self.options.keys())
+        # Catch deprecated options
```
smite, line 94: Is there a deprecation warning given to the user somewhere?

smite (approved):

PR #66, opened 2023-06-06
Add IQMFakeApollo

JMuff22:
Hi, this pull request adds `IQMFakeApollo` to provide noisy simulation of IQM's Apollo chip. I created the changes based off `IQMFakeAdonis`. The values are currently placeholder values which are open to be changed by the maintainers. 

Should any documentation be added regarding this PR? Let me know and I will add it.

PR #113, opened 2024-08-14
Add iqm fake aphrodite

JMuff22:
Hi, this pull request adds `IQMFakeAphrodite` to provide noisy simulation of IQM's Aphrodite chip. I created the changes based off IQMFakeApollo. The values are currently placeholder values which are open to be changed by the maintainers. They are based off https://github.com/iqm-finland/qiskit-on-iqm/pull/66. 

Should any documentation be added regarding this PR? Let me know and I will add it.

Additionally, I added 1 comment to `CONTRIBUTING.rst` to inform contributors that they can run the formatting through tox. 

**Note that the tests will fail due to duplicate code as the values are copied from IQM fake apollo**


Aerylia (approved):
If you can rename the noise model name, before merging, that would be nice. But the other fake backends don't have a sensible name either, so it's not a big deal.

JMuff22:
> You'll have to fix the linting issues for tests to pass.
> 
> If you can rename the noise model name, before merging, that would be nice. But the other fake backends don't have a sensible name either, so it's not a big deal.

@Aerylia  It was my understanding that the CI test fails due to "duplicate code" between the noise model for aphrodite and apollo. This is because the values themselves are copy and pasted. I am not sure how to fix this issue other than use different values. The values are placeholders anyway if we follow the convention from https://github.com/iqm-finland/qiskit-on-iqm/pull/66

PR #114, opened 2024-08-21
Adjustments needed to support Qiskit v1

Aerylia:
Some minor changes in implementation details of the NaiveMoveTranspiler.
Changed usage of `execute` in tests to `transpile` followed by `backend.run`.
Adjusted the method of iterating through a QuantumCircuit for serialisation and testing. To be compatible with the next major Qiskit release.

Tests generate some Deprecation warnings. These come from Qiskit Aer. Except for 1 warning raised in the NaiveMoveTranspiler that should be resolved with COMP-1361.

Remaining tasks:
- [x] Update the user guide documentation
- [x] Update Changelog
- [x] Update requirements/tox
- [x] ~~Remove use of `BackendProperties` from `transpile_to_iqm` to support `qiskit==1.2`~~ Later


pyproject.toml
```diff
@@ -20,8 +20,8 @@ classifiers = [
 requires-python = ">=3.9, <3.12"
 dependencies = [
     "numpy",
-    "qiskit >= 0.45.1, < 0.46",
-    "qiskit-aer >= 0.13.1, < 0.14",
+    "qiskit >= 1.0",
+    "qiskit-aer >= 0.14",
```
Aerylia, line 24: This needs some upper bound.

src/iqm/qiskit_iqm/iqm_move_layout.py
```diff
@@ -139,8 +139,8 @@ def _require_type(qubit_index: int, required_type: str, instruction_name: str):
                 # The move operation requires that the first operand is the move qubit,
                 # and the second must be the resonator
                 (qubit, resonator) = node.qargs
-                _require_type(qubit.index, 'move_qubit', 'move')
-                _require_type(resonator.index, 'resonator', 'move')
+                _require_type(qubit._index, 'move_qubit', 'move')
+                _require_type(resonator._index, 'resonator', 'move')
```
smite, line 143: Maybe we should not use private attributes. Is there some other, recommended way of doing this?

smite: Elsewhere as well.

tests/test_iqm_provider.py
```diff
@@ -113,10 +113,18 @@ def test_qubit_name_to_index_to_qubit_name(adonis_architecture_shuffled_names):
...
+    client = IQMClient(url="http://some_url")
+    client._token_manager = None  # Do not use authentication
+    when(client).get_quantum_architecture().thenReturn(linear_architecture_3q)
+    expect(requests, times=1).post(client._base_url, ANY).thenReturn(
+        MockJsonResponse(201, {'id': str(uuid.UUID('3c3fcda3-e860-46bf-92a4-bcc59fa76ce9'))})
+    )
+    backend = IQMBackend(client)
+    circuit = QuantumCircuit(3)
+    circuit.cz(0, 2)
+    with pytest.raises(CircuitExecutionError):
```
smite, line 126: Should have a `match` with some error text.

smite:
The User Guide (in docs) needs to be updated as well.

src/iqm/qiskit_iqm/iqm_move_layout.py
Aerylia: Try to find these instances and remove them if trivial. In cases where this cannot be simply fixed, I'll make a note and discussion point.

Aerylia: I managed to remove all instances. In some cases, this might cause a small performance hit for looking up the index of a qubit from the circuit, but all those cases can be removed in the near future as part of the traspiler improvements ticket COMP-1361.

CHANGELOG.rst
```diff
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+Version 13.11
+=============
+
+* Adjustments needed to support Qiskit V1 that are backwards compatible with ``qiskit < 1.0``. `#114 <https://github.com/iqm-finland/qiskit-on-iqm/pull/114>`_
+* Updated Qiskit dependencies and testing to support ``qiskit >= 0.45.3 < 1.2`` and ``qiskit-aer >= 0.13 < 0.15``. 
+* Adjusted documentation to recommend the use to :meth:`qiskit.transpile()` or :meth:`transpile_to_IQM()` in combination with :meth:`backend.run()` instead of using :meth:`execute()`.
```
jkotilahti, line 10:
```suggestion
* Adjusted documentation to recommend the use of :meth:`qiskit.transpile()` or :meth:`transpile_to_IQM()` in combination with :meth:`backend.run()` instead of using :meth:`execute()`.
```
or just "using".

pyproject.toml
```diff
@@ -20,8 +20,8 @@ classifiers = [
 requires-python = ">=3.9, <3.12"
 dependencies = [
     "numpy",
-    "qiskit >= 0.45.1, < 0.46",
-    "qiskit-aer >= 0.13.1, < 0.14",
+    "qiskit >= 0.45, < 1.2",
```
jkotilahti, line 23: Should we allow also Qiskit 1.2? At least it's already included in the test environments.

tests/fake_backends/test_fake_deneb.py
```diff
@@ -159,7 +171,7 @@ def test_execute_works_but_backend_run_doesnt_with_unsupported_gates():
     qc_list[3].z(4)
 
     for qc in qc_list:
-        execute(qc, backend, shots=1000)
+        backend.run(transpile_to_IQM(qc, backend=backend), shots=1000)
```
jkotilahti, line 174: The name of this test `test_execute_works_but_backend_run_doesnt_with_unsupported_gates` is now outdated, could be renamed to something like `test_backend_run_doesnt_work_with_unsupported_gates`.

tests/test_iqm_backend.py
```diff
@@ -65,7 +65,10 @@ def test_transpile(backend):
 
     circuit_transpiled = transpile(circuit, backend=backend)
     cmap = backend.coupling_map.get_edges()
-    for instruction, qubits, _ in circuit_transpiled.data:
+    print(circuit_transpiled.data)
```
jkotilahti, line 68:
```suggestion
```
I'm guessing this was just for debugging and can be removed.

tests/test_iqm_provider.py
```diff
@@ -121,10 +129,20 @@ def test_qubit_name_to_index_to_qubit_name(adonis_architecture_shuffled_names):
...
-    circuit = QuantumCircuit(2, 2)
-    with pytest.raises(ValueError, match='has not been transpiled against the current backend'):
-        backend.serialize_circuit(circuit)
+def test_serialize_circuit_raises_error_for_non_transpiled_circuit(circuit, linear_architecture_3q):
+    client = IQMClient(url='http://some_url')
+    client._token_manager = None  # Do not use authentication
+    when(client).get_quantum_architecture().thenReturn(linear_architecture_3q)
+    expect(requests, times=1).post(client._base_url, ANY).thenReturn(
+        MockJsonResponse(201, {'id': str(uuid.UUID('3c3fcda3-e860-46bf-92a4-bcc59fa76ce9'))})
+    )
```
jkotilahti, line 138: The `expect(requests, times=1)` does not actually check that it's called 1 time unless there is also `verifyNoUnwantedInteractions()` at the end of the test (https://mockito-python.readthedocs.io/en/latest/the-functions.html#mockito.expect). If that is added, you can see that this request is in fact never sent during this test, because the error is raised before that. I would suggest just removing this part (and MockJsonResponse class since it's not used elsewhere).

```suggestion
```

tests/test_iqm_provider.py
```diff
@@ -541,7 +544,7 @@ def test_create_run_request(backend, circuit, create_run_request_default_kwargs,
     when(backend.client).submit_run_request(run_request).thenReturn(uuid.uuid4())
 
     assert backend.create_run_request(circuit_transpiled, **options) == run_request
```
jkotilahti, line 546: Can remove `options` also from here (since it just has transpilation options), I forgot it after the last change in my PR.

```suggestion
    assert backend.create_run_request(circuit_transpiled) == run_request
```

pyproject.toml
Aerylia: It does not work because of the deprecation of BackendProperties that is used in the Deneb transpiler. Ive added qiskit v1.2 support to the transpiler improvements ticket so these changes can be made sequentially.

tests/fake_backends/test_fake_deneb.py
Aerylia: Renamed it to `test_transpiling_works_but_backend_run_doesnt_with_unsupported_gates`

pyproject.toml
jkotilahti: Then I think `qiskit-v1.2` should be removed from `tox.ini` environments for now.

tests/fake_backends/test_fake_deneb.py
jkotilahti: (looks like it's not actually renamed yet)

jkotilahti (approved):
Looks good!

There is still the minor comment about renaming one test above, but I approve already the actual changes.

(Also when merging, I would recommend clicking "squash and merge" instead of just "merge" to keep `main` commit history cleaner, without unnecessary intermediate commits like `pylint fixes`. Although I guess this has not been followed consistently by everyone.)

Aerylia:
Forgot to save the file <_<

jkotilahti (approved):

Issue #118, opened 2024-09-12
`optimize_single_qubit_gates` does not preserve layout


JMuff22:
When using the function `optimize_single_qubit_gates` on an already transpiled circuit (as per the docs), the layout provided to the transpiler is not preserved. The means that an additional transpiler pass is needed after running `optimize_single_qubit_gates`. 

## Reproducible example

```
from qiskit import QuantumCircuit
from qiskit import transpile
from iqm.qiskit_iqm import IQMFakeAdonis
from iqm.qiskit_iqm.iqm_transpilation import optimize_single_qubit_gates

qc = QuantumCircuit(3)
qc.h(0)
qc.cx(0, 1)
qc.cx(0, 2)
qc.measure_all()

print(qc.draw(output='text'))

backend = IQMFakeAdonis()
transpiled_circuit = transpile(qc, backend=backend)

print()
print(transpiled_circuit.draw(output='text'))

qc_optimized = optimize_single_qubit_gates(transpiled_circuit)

print()
print(qc_optimized.draw(output='text', idle_wires=False))
```

Gives 

```
        ┌───┐           ░ ┌─┐      
   q_0: ┤ H ├──■────■───░─┤M├──────
        └───┘┌─┴─┐  │   ░ └╥┘┌─┐   
   q_1: ─────┤ X ├──┼───░──╫─┤M├───
             └───┘┌─┴─┐ ░  ║ └╥┘┌─┐
   q_2: ──────────┤ X ├─░──╫──╫─┤M├
                  └───┘ ░  ║  ║ └╥┘
meas: 3/═══════════════════╩══╩══╩═
                           0  1  2 
layout=None

global phase: π/2
               ┌────────────┐┌────────┐   ┌────────────┐  ┌────────┐             ░    ┌─┐   
      q_1 -> 0 ┤ R(π/2,π/2) ├┤ R(π,0) ├─■─┤ R(π/2,π/2) ├──┤ R(π,0) ├─────────────░────┤M├───
               └────────────┘└────────┘ │ └────────────┘  └────────┘             ░    └╥┘   
ancilla_0 -> 1 ─────────────────────────┼──────────────────────────────────────────────╫────
               ┌────────────┐┌────────┐ │                                        ░ ┌─┐ ║    
      q_0 -> 2 ┤ R(π/2,π/2) ├┤ R(π,0) ├─■───────■────────────────────────────────░─┤M├─╫────
[...]

PR #116, opened 2024-09-05
Transpiler integration refactor and qiskit v1.2 support

Aerylia:
Refactor of ``transpile_to_IQM`` to support qiskit v1.2.
Part of COMP-1361

Task list:
- [x] Remove the use of Qiskit's BackendProperties
- [x] Update tox and pyproject files for the new version
- [x] Create a simplified and a real target in `IQMBackendBase` to use instead of BackendProperties
- [x]  Fix the transpile layout preservation in transpile_to_IQM. It is currently ignored. SW-601 and IQMCS-159
- [x] Add unittests to check if the layout is properly preserved.
- [x] Add the option to restrict transpilation to a set of connected qubits
- [x] Update documentation
- [x] Check e2e tests
- [x] Discuss version bump. There is a major backward compatible workflow change (native Qiskit transpile integration) and a small breaking API change for the IQMBackend, so major release it is.

Other changes:
- Refactored `create_run_request` to improve the experience when using different run options, by making them explicit arguments that our documentation can refer to
- Updated the user guide to refer to the `create_run_request ` documentation to avoid out-dated duplicated information.
- Adjusted `qubit_name_to_idx` and `idx_to_qubit_name` in `IQMBackendBase` to no longer return Optional, but raise an error when the name or index cannot be found. This improved mypy and pylint issues as well as user experience since these functions should never be called in a way that it would return None.
- Native transpiler integration
- Addition of transpiler plugins
- Deneb bugfix.
- Drop Python 3.9 support
- Drop Qiskit < 1.0 support
- Refactor the `iqm_move_layout ` module

PR #119, opened 2024-09-12
Preserve layout in `optimize_single_qubit_gates`.

JMuff22:
Fixes https://github.com/iqm-finland/qiskit-on-iqm/issues/118 

Currently, ancilla qubits are not added to the circuit. So the number of qubits in the quantum circuit is still reduced.

### DIFF
--- a/tests/test_iqm_transpilation.py
+++ b/tests/test_iqm_transpilation.py
@@ -127,3 +127,22 @@ def test_submitted_circuit(adonis_architecture):
         'measure:2',
         'measure:4',
     ]
+
+
+def test_optimize_single_qubit_gates_preserves_layout():
+    circuit = QuantumCircuit(3)
+    circuit.h(0)
+    circuit.cx(0, 1)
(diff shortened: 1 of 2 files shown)


JMuff22:
Additionally, the file `src/iqm/qiskit-on-iqm/iqm_transpilation.py` has additional problems:
1. Usage of single quotes. This project uses black which prefers double quotes
2. Inconsistent British and american english `optimise` vs `optimize`. Also `neighbouring`

tests/test_iqm_transpilation.py
Aerylia, line 1: This is a good start. Could you add some different cases? e.g. where the transpiler generates the initial layout or generates a final layout so that those are preserved as well?

Aerylia (changes requested):

Aerylia:
Thank you for taking this on!

Aerylia:
Will be obsolete with #116

PR #122, opened 2024-10-04
Remove unnecessary build files when publishing documentation

jkotilahti:
Does the same thing as https://github.com/iqm-finland/iqm-client/pull/133.

Here I didn't add documentation about the alternative `git clone` options, because the repository size is not as big as iqm-client (~50MB vs ~500MB), so it's still pretty fast to clone normally.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -69,6 +69,8 @@ jobs:
         run: |
           tox -e docs
           touch build/sphinx/html/.nojekyll  # allow underscores in URL path
+          # remove unnecessary build files
+          sudo rm -rf build/sphinx/html/.doctrees
       - name: Publish to gh-pages
         uses: JamesIves/github-pages-deploy-action@4.1.4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -69,6 +69,8 @@ jobs:
         run: |
           tox -e docs
           touch build/sphinx/html/.nojekyll  # allow underscores in URL path
+          # remove unnecessary build files
+          sudo rm -rf build/sphinx/html/.doctrees
```
rbrazinskas, line 73: Could we achieve the same effect by updating `.gitignore` file with appropriate entry?

jkotilahti: I tried it initially for iqm-client ([commit 5de3df2](https://github.com/jkotilahti/iqm-client/commit/332531eafafd23c81f5b47e101894364b70a4d59) in https://github.com/jkotilahti/iqm-client/commits/main/), but it didn't work.

rbrazinskas (approved):

PR #123, opened 2024-10-15
Add support for classically controlled PRX gates.

smite:
* Adds support for classically controlled R, X, Y, RX, RY gates.
* Adds support for the Reset instruction, in a somewhat hacky way.
* Remove the deprecated `phased_rx` and `measurement` native ops.
* Split the provider and backend tests in two files.


Aerylia (commented):
Are there any tests that check if the `c_if` is working as intended? It seems that everything is handled Qiskit side, which is good, but we should have some tests that will break if Qiskit breaks itself.

I'm also not seeing any code that would make the Qiskit circuit into the appropriate IQM Circuit.

docs/user_guide.rst
```diff
@@ -233,6 +233,56 @@ time you do not need to deal with IQM-style qubit names when using Qiskit, howev
...
 
+Classically controlled gates
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some IQM backends support classically controlled gates, but this support currently has several limitations:
+
+* Only the ``x``, ``y``, ``rx``, ``ry`` and ``r`` gates can be classically controlled.
+* The gates can only conditioned on one classical bit, and the only control available is to
+  apply the gate if the bit is 1, and apply an indentity gate if the bit is 0.
+* The availability of the controlled gates depends on the instrumentation of the backend.
```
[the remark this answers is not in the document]
smite: The information will be in the DQA.

docs/user_guide.rst
```diff
@@ -233,6 +233,56 @@ time you do not need to deal with IQM-style qubit names when using Qiskit, howev
...
+
+    from qiskit import QuantumCircuit
+
+    qr = QuantumRegister(2, 'q')
+    cr = ClassicalRegister(1, 'c')
+    circuit = QuantumCircuit(qr, cr)
+
+    circuit.h(0)
+    circuit.measure(0, cr[0])
+    circuit.x(1).c_if(cr, 1)
```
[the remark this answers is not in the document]
smite: c_if always uses the full register, and interprets it as an integer AFAIK. This is why we need to use 1-bit classical registers with the classical control.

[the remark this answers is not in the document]
smite: See the note below.

Aerylia:
Can you update the changes with main? The review would become invalid afterwards anyways and it'll make it easier to see what I did and didn't review yet.

docs/user_guide.rst
```diff
@@ -243,6 +243,56 @@ time you do not need to deal with IQM-style qubit names when using Qiskit, howev
 :meth:`~.IQMBackendBase.qubit_name_to_index` and :meth:`~.IQMBackendBase.index_to_qubit_name` can become handy.
 
 
+Classically controlled gates
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some IQM backends support classically controlled gates, but this support currently has several limitations:
```
smite, line 249:
```suggestion
Some IQM quantum computers support classically controlled gates, that is, gates that are executed conditionally depending on the result of a measurement preceding them in the quantum circuit. This support currently has several limitations:
```

docs/user_guide.rst
```diff
@@ -243,6 +243,56 @@ time you do not need to deal with IQM-style qubit names when using Qiskit, howev
...
+
+* Only the ``x``, ``y``, ``rx``, ``ry`` and ``r`` gates can be classically controlled.
+* The gates can only be conditioned on one classical bit, and the only control available is to
+  apply the gate if the bit is 1, and apply an indentity gate if the bit is 0.
+* The availability of the controlled gates depends on the instrumentation of the backend.
```
smite, line 254:
```suggestion
* The availability of the controlled gates depends on the instrumentation of the quantum computer.
```

docs/user_guide.rst
```diff
@@ -243,6 +243,56 @@ time you do not need to deal with IQM-style qubit names when using Qiskit, howev
...
+       q_1: ──────╫──┤ X ├──░──╫─┤M├
+                  ║  └─╥─┘  ░  ║ └╥┘
+                  ║ ┌──╨──┐    ║  ║
+       c: 1/══════╩═╡ 0x1 ╞════╬══╬═
+                  0 └─────┘    ║  ║
+    meas: 2/═══════════════════╩══╩═
+                               0  1
+
+
+Executing the above circuit should result in the counts being approximately 50/50 split
```
smite, line 287:
```suggestion
The first measurement operation stores its result in the 1-bit classical register ``c``. If the result is 1, the ``X`` gate will be applied. If it is zero, an identity gate of corresponding duration is applied instead.

Executing the above circuit should result in the counts being approximately 50/50 split
```

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -240,49 +241,103 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
+                for q in qubit_names:
+                    physical_qubit_name = self._idx_to_qb[int(q)]
+                    instructions.append(
+                        Instruction(
+                            name='cc_prx',
+                            qubits=[q],
+                            args={
+                                'angle_t': 0.5,
+                                'phase_t': 0.0,
+                                'feedback_key': feedback_key,
```
Aerylia, line 301: How does the QPU know which bit in the feedback_key measurement belongs to which qubit? We measure n qubits and cc_prx on 1 qubit n times. 
If you are not allowed to control from a measurement on another qubit, the documentation example would not work.

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -240,49 +241,103 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
+                    raise ValueError(
+                        'This backend only supports conditionals on r, x, y, rx and ry gates,'
+                        f' not on {instruction.name}'
+                    )
+                native_inst.name = 'cc_prx'
+                creg, value = condition
+                if len(creg) != 1:
+                    raise ValueError(f'{instruction} is conditioned on multiple bits, this is not supported.')
+                if value != 1:
+                    raise ValueError(f'{instruction} is conditioned on integer value {value}, only 1 is supported.')
```
Aerylia, line 327:
```suggestion
                    raise ValueError(f'{instruction} is conditioned on integer value {value}, only value 1 is supported.')
```

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -240,49 +241,103 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
             elif instruction.name == 'move':
-                instructions.append(Instruction(name='move', qubits=qubit_names, args={}))
+                native_inst = Instruction(name='move', qubits=qubit_names, args={})
             elif instruction.name == 'barrier':
-                instructions.append(Instruction(name='barrier', qubits=qubit_names, args={}))
+                native_inst = Instruction(name='barrier', qubits=qubit_names, args={})
             elif instruction.name == 'measure':
-                mk = MeasurementKey.from_clbit(clbits[0], circuit)
-                instructions.append(Instruction(name='measure', qubits=qubit_names, args={'key': str(mk)}))
+                clbit = circuit_instruction.clbits[0]  # always a single-qubit measurement
```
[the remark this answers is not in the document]
smite: Added a check.

src/iqm/qiskit_iqm/iqm_provider.py
smite: `cc_prx` takes both a `feedback_key` and a `feedback_qubit`. The key indentifies the measurement op (there may be multiple), and the qubit identifies the qubit.

Aerylia (commented):
I commented some questions, but otherwise it looks good.

src/iqm/qiskit_iqm/iqm_provider.py
Aerylia: Ah I see. So if I wanted to mix up the reset measurements and their X gates, I would shuffle the `feedback_qubit`s not the `feedback_key`. That makes sense.

smite:
> Are there any tests that check if the `c_if` is working as intended? It seems that everything is handled Qiskit side, which is good, but we should have some tests that will break if Qiskit breaks itself.
> 
> I'm also not seeing any code that would make the Qiskit circuit into the appropriate IQM Circuit.

There are unit test in test_iqm_backend.py, and there will be E2E tests.

Aerylia (approved):

Aerylia (approved):

Aerylia (approved):

PR #124, opened 2024-10-25
Use dynamic quantum architecture (DQA) as transpilation target

jkotilahti:
- IQMBackend gets DynamicQuantumArchitecture instead of QuantumArchitectureSpecification from client.

- IQMBackendBase supports creation only from DynamicQuantumArchitecture instead of QuantumArchitectureSpecification .

- IQMFakeBackend still uses the old QuantumArchitectureSpecifications, since it doesn't have an associated calibration set. It converts the QuantumArchitectureSpecification into a "fake" DynamicQuantumArchitecture that is passed to IQMBackendBase. 

- `IQMBackend.run()` warns the user if default calibration set changed, and also adds a more helpful message when encountering CircuitValidationError.

- @smite I copied some code from https://github.com/iqm-finland/qiskit-on-iqm/pull/123 which was required to make the tests pass

- Fixes PyPI publishing in the same way as https://github.com/iqm-finland/iqm-client/pull/143 by @rbrazinskas


src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -16,81 +16,117 @@
...
+@dataclass
+class QuantumArchitecture:
+    """A representation of the quantum computer architecture.
+
+    Used to represent data from a DynamicQuantumArchitecture or a QuantumArchitectureSpecification in a unified way.
+    """
+
+    qubits: list[str]
+    computational_resonators: list[str]
+    gate_loci: dict[str, tuple[Locus, ...]]
```
smite, line 43: How about just using DynamicQuantumArchitecture instead of this one, and adding a function for creating a fake DQA from a static architecture just for FakeBackends?
You could create a single fake implementation for each gate, containing all the loci.

smite: And put something like the all-zeros UUID in the UUID field. It's a hack but it should be safe.

jkotilahti: Yeah, that's another option I was considering. I did it this way because there doesn't seem to be a clean way to indicate a "fake"/"dummy"/"unused" UUID, but I guess the all-zeros one would be ok. For me either way is fine.

jkotilahti: Removed QuantumArchitecture now and using a "fake" DQA instead.

src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -16,81 +16,104 @@
...
+    GateInfo,
+    QuantumArchitectureSpecification,
+)
 from iqm.qiskit_iqm.move_gate import MoveGate
 
 IQM_TO_QISKIT_GATE_NAME: Final[dict[str, str]] = {'prx': 'r', 'cz': 'cz'}
 
 
+def dqa_from_static_architecture(static_architecture: QuantumArchitectureSpecification) -> DynamicQuantumArchitecture:
+    """Returns a dynamic quantum architecture created from the given static architecture."""
```
smite, line 40:
```suggestion
    """Returns a dynamic quantum architecture (with mock gate implementations) created from the given static architecture."""
```

src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -16,81 +16,104 @@
...
+    GateImplementationInfo,
+    GateInfo,
+    QuantumArchitectureSpecification,
+)
 from iqm.qiskit_iqm.move_gate import MoveGate
 
 IQM_TO_QISKIT_GATE_NAME: Final[dict[str, str]] = {'prx': 'r', 'cz': 'cz'}
 
 
+def dqa_from_static_architecture(static_architecture: QuantumArchitectureSpecification) -> DynamicQuantumArchitecture:
```
smite, line 39:
```suggestion
def _dqa_from_static_architecture(static_architecture: QuantumArchitectureSpecification) -> DynamicQuantumArchitecture:
```

src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -16,81 +16,104 @@
...
-    def __init__(self, architecture: QuantumArchitectureSpecification, **kwargs):
+    def __init__(
+        self,
+        architecture: Union[QuantumArchitectureSpecification, DynamicQuantumArchitecture],
+        **kwargs,
+    ):
         super().__init__(**kwargs)
-        self.architecture = architecture
+        if isinstance(architecture, QuantumArchitectureSpecification):
+            arch = dqa_from_static_architecture(architecture)
```
smite, line 75:
```suggestion
            arch = _dqa_from_static_architecture(architecture)
```

src/iqm/qiskit_iqm/iqm_job.py
```diff
@@ -167,7 +170,7 @@ def result(self) -> Result:
             self._result = self._format_iqm_results(results)
             # RemoteIQMBackend.run() populates circuit_metadata, so it may be None if method wasn't called in current
             # session. In that case retrieve circuit metadata from RunResult.metadata.request.circuits[n].metadata
-            if self.circuit_metadata is None:
+            if self.circuit_metadata is None and results.metadata.request is not None:
```
smite, line 173: Why would the request not be there?

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -284,13 +305,13 @@ class IQMFacadeBackend(IQMBackend):
...
         self.fake_adonis = IQMFakeAdonis()
-        target_architecture = client.get_quantum_architecture()
+        target_architecture = client.get_dynamic_quantum_architecture(kwargs.get('calibration_set_id', None))
 
         if not self.fake_adonis.validate_compatible_architecture(target_architecture):
             raise ValueError('Quantum architecture of the remote quantum computer does not match Adonis.')
 
         super().__init__(client, **kwargs)
-        self.name = f'IQMFacade{target_architecture.name}Backend'
+        self.name = 'IQMFacadeBackend'
```
Aerylia, line 314: DQA does not contain a name?
I've made a bug report that the facade only works for Adonis, because there is not reason for that.

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -338,13 +359,17 @@ def __init__(self, url: str, **user_auth_args):  # contains keyword args auth_se
...
+    ) -> Union[IQMBackend, IQMFacadeBackend]:
         """An IQMBackend instance associated with this provider.
 
         Args:
             name: optional name of a custom facade backend
+            calibration_set_id: ID of the calibration set used to create the transpilation target of the backend.
+                If None, the server default calibration set will be used.
         """
         client = IQMClient(self.url, client_signature=f'qiskit-iqm {__version__}', **self.user_auth_args)
         if name == 'facade_adonis':
```
Aerylia, line 373: Maybe this should be the name for the IQM facade backend.

tests/fake_backends/test_iqm_fake_backend.py
Aerylia, line 1: I guess the reason that this is using the old QA instead of the DQA is because the FakeBackend still uses it. Is there any reason why that is needed? If we want to deprecate the old QA later, we need to rewrite this again. Might as well do that now.

tests/move_architecture/test_architecture.py
```diff
@@ -86,10 +49,10 @@ def test_backend_configuration_adonis(adonis_architecture):
     assert backend.target.physical_qubits == [0, 1, 2, 3, 4]
     assert set(backend.target.operation_names) == {'r', 'id', 'cz', 'measure'}
     assert [f'{o.name}:{o.num_qubits}' for o in backend.target.operations] == [
```
Aerylia, line 51: Is the ordering here deterministic? Why did the ordering change?

Aerylia (changes requested):
Looks good. I think the DQA should also be adapted into the FakeBackend because that will make the task of deprecating the old QA trivial.

tests/move_architecture/test_architecture.py
jkotilahti: It is deterministic, it comes from the order in which instructions are added to `target` in IQMBackendBase. The changes related to that were taken from Ville's https://github.com/iqm-finland/qiskit-on-iqm/pull/123.

src/iqm/qiskit_iqm/iqm_provider.py
jkotilahti: Yes, DQA does not contain a name, only a calset id.

src/iqm/qiskit_iqm/iqm_provider.py
jkotilahti: I think it's intentionally `facade_adonis`, it's documented in User Guide. I would not change it here.

docs/user_guide.rst
```diff
@@ -133,14 +133,21 @@ circuit(s) are sampled:
...
-For example, if you know an ID of a specific calibration set that you want
-to use, you can provide it as follows:
+For example, you can enable heralding measurements by passing the appropriate circuit compilation option as follows:
 
 .. code-block:: python
 
-    job = backend.run(transpiled_circuit, shots=1000, calibration_set_id="f7d9642e-b0ca-4f2d-af2a-30195bd7a76d")
+    from iqm.iqm_client import CircuitCompilationOptions
+
+    job = backend.run(transpiled_circuit, shots=1000, circuit_compilation_options=CircuitCompilationOptions(heralding_mode=HeraldingMode.ZEROS))
```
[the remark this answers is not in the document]
jkotilahti: `backend.run()` now always uses the calibration set of the `backend`, so using the latest one may require creating a new `backend`. A warning is shown in that case, see `"Server default calibration set has changed ..."` warning in `IQMBackend.create_run_request`, which suggests to use a new backend. I'll also add a mention about this in the user guide as suggested in your other comment above.

docs/user_guide.rst
```diff
@@ -133,14 +133,21 @@ circuit(s) are sampled:
    `Inspecting circuits before submitting them for execution`_ for inspecting the actual run request sent for
    execution.
 
+The calibration data for an IQM quantum computer is stored in a calibration set. An :class:`.IQMBackend` instance
+always corresponds to a specific calibration set, so that its transpilation target uses only those QPU components
+(qubits and computational resonators) and gates which are available in that calibration set. The server default
+calibration set will be used by default, but one can also use a different calibration set by specifying the
+``calibration_set_id`` parameter for :meth:`.IQMProvider.get_backend` or :class:`.IQMBackend`.
+
```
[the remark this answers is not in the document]
jkotilahti: Added explanation.

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -50,11 +57,16 @@ class IQMBackend(IQMBackendBase):
     """
 
     def __init__(self, client: IQMClient, **kwargs):
-        architecture = client.get_quantum_architecture()
+        # TODO: using pop() instead of get() because passing non-empty kwargs to BackendV2 leads to an error
```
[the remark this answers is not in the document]
jkotilahti: Done.

(looks like Ville already did this change in https://github.com/iqm-finland/qiskit-on-iqm/pull/123, so I copied the changes from there)

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -50,11 +57,16 @@ class IQMBackend(IQMBackendBase):
...
+        #  which looks like a Qiskit bug.
+        calset_id = kwargs.pop('calibration_set_id', None)
+        self._use_default_calibration_set = calset_id is None
+        architecture = client.get_dynamic_quantum_architecture(calset_id)
         super().__init__(architecture, **kwargs)
         self.client: IQMClient = client
         self._max_circuits: Optional[int] = None
-        self.name = f'IQM{architecture.name}Backend'
+        self.name = 'IQM Backend'
+        self._calibration_set_id = architecture.calibration_set_id
```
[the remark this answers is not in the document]
jkotilahti: Done.

src/iqm/qiskit_iqm/iqm_provider.py
jkotilahti: Changed since only Adonis is supported for facade backend.

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -176,13 +182,28 @@ def create_run_request(self, run_input: Union[QuantumCircuit, list[QuantumCircui
...
+        try:
+            run_request = self.client.create_run_request(
+                circuits_serialized,
+                qubit_mapping=qubit_mapping,
+                calibration_set_id=self._calibration_set_id,
+                shots=shots,
+                options=merged_options['circuit_compilation_options'],
+            )
+        except CircuitValidationError as e:
+            raise CircuitValidationError(
```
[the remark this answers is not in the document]
jkotilahti: Changed the wording slightly to cover also the case where circuits were also not transpiled at all. We don't have detailed information about the exact cause from iqm-client, but the message should now cover all cases where CircuitValidationError can be thrown:
- circuits not transpiled at all
- circuits transpiled using different backend than used to submit them

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -176,13 +182,28 @@ def create_run_request(self, run_input: Union[QuantumCircuit, list[QuantumCircui
...
         qubit_mapping = {str(idx): qb for idx, qb in self._idx_to_qb.items() if idx in used_indices}
 
-        return self.client.create_run_request(
-            circuits_serialized,
-            qubit_mapping=qubit_mapping,
-            calibration_set_id=calibration_set_id if calibration_set_id else None,
-            shots=shots,
-            options=merged_options['circuit_compilation_options'],
-        )
+        if self._use_default_calibration_set:
```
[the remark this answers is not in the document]
jkotilahti: IQMBackend now always corresponds to a specific calibration set, which is always used to run the circuits, and thus `create_run_request` does not allow specifying a calset id. There are the following ways to create a backend: 

- `IQMBackend(client)` -> the backend will always use the calibration set that was default at the time when backend was created

- `IQMBackend(client, calibration_set_id=custom_calset)` -> the backend will always use `custom_calset` 

Using a new server default calset requires creating a new backend, as now explained in documentation. I think no change is needed here.

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -28,7 +28,7 @@
 from qiskit_aer.noise import NoiseModel
 from qiskit_aer.noise.errors import depolarizing_error, thermal_relaxation_error
 
-from iqm.iqm_client import QuantumArchitectureSpecification
+from iqm.iqm_client import DynamicQuantumArchitecture, QuantumArchitectureSpecification
```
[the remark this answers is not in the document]
jkotilahti: The reasons why I still used QuantumArchitectureSpecification for FakeBackend are:

* Conceptually, FakeBackend does not have a calibration set, while a DynamicQuantumArchitecture always corresponds to a calibration set.

* More convenient to create the FakeBackends, since there is no need to add fake gate implementation info and fake calset ids when defining them. It makes their definitions cleaner since there is not this extra info which would be unused anyway.

* Keeping it using QuantumArchitectureSpecification requires less code changes, less changes to review.

* In the follow-up issue SW-707 where we change how static architectures work, there is a note about the possibility to use the static/idealistic arhcitecture to create the simulated backends. If we change them now to use DQA, we might need to change again to use the new SQA (so we change `old SQA -> DQA -> new SQA`), while possibly it would be enough to just do the change then (`old SQA -> new SQA`).

Because of these reasons, I would still keep FakeBackend using QuantumArchitectureSpecification, and revisit it in the follow-up issue.

tests/fake_backends/test_iqm_fake_backend.py
jkotilahti: See the other thread about FakeBackend and DQA, let's continue possible conversation there.

src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -78,18 +41,13 @@ class IQMBackendBase(BackendV2, ABC):
 
     def __init__(
         self,
-        architecture: Union[QuantumArchitectureSpecification, DynamicQuantumArchitecture],
+        architecture: DynamicQuantumArchitecture,
```
Aerylia, line 44: I think this will be problematic for people running Qiskit on IQM with on devices with older software. We should probably support both.

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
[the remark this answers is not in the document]
Aerylia: Not allowing SQA for IQMBackendBase makes the library unusable for devices that don't support DQA yet. So that's not good. What I'm saying is that the IQMFakeBackend should support both as well, because we need have the SQA available for some of these devices right now, and we want to use it for creating an idealistic fake backend. But sometimes you want an accurate fake backend, for which you also need the DQA.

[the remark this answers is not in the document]
jkotilahti: I see, I reverted the commit which removed SQA support from IQMBackendBase. But I would still not add DQA support to IQMFakeBackend, because we currently only support hardcoded FakeBackends, and the above reasons are valid considering. I think IQMFakeBackend support for DQA should be done in the issue which actually adds support for IQMFakeBackends using real calibration data.

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -106,7 +112,7 @@ def __init__(
         name: str = "IQMFakeBackend",
         **kwargs,
     ):
-        super().__init__(architecture, **kwargs)
+        super().__init__(self._dqa_from_static_architecture(architecture), **kwargs)
```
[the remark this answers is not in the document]
jkotilahti: I think IQMFakeBackend support for DQA should be done in that issue which actually adds support for IQMFakeBackends using real calibration data, so far we only support hardcoded fake devices.

Aerylia (approved):

docs/user_guide.rst
```diff
@@ -133,14 +133,24 @@ circuit(s) are sampled:
...
    execution.
 
+The calibration data for an IQM quantum computer is stored in a calibration set. An :class:`.IQMBackend` instance
+always corresponds to a specific calibration set, so that its transpilation target uses only those QPU components
+(qubits and computational resonators) and gates which are available in that calibration set. The server default
+calibration set will be used by default, but you can also use a different calibration set by specifying the
+``calibration_set_id`` parameter for :meth:`.IQMProvider.get_backend` or :class:`.IQMBackend`. If the server default
+calibration set has changed after you created the backend, the backend would still use the original default calibration
+set when submitting circuits. In this case, a warning will be shown to notify you about this, and you would need to
+create a new backend if you want to use the newer calibration set.
```
smite, line 143:
```suggestion
calibration set has changed after you have created the backend, the backend will still use the original default calibration
set when submitting circuits for execution. In this case, a warning will notify you about this.
You will need to create a new backend if you want to use the new default calibration set.
```

jkotilahti: Sorry, merged already, can be done in https://github.com/iqm-finland/qiskit-on-iqm/pull/123 I guess.

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
```diff
@@ -370,3 +370,20 @@ def run(self, dag):
         job = sim_noise.run(circuits, shots=shots)
 
         return job
+
+    def validate_compatible_architecture(self, architecture: DynamicQuantumArchitecture) -> bool:
+        """Given a dynamic quantum architecture returns true if its number of components, names of components and
+        component connectivity matches the architecture of this backend."""
```
smite, line 376:
```suggestion
        """Compare a dynamic quantum architecture to the static architecture of the fake backend.

        Args:
            architecture: dynamic quantum architecture to compare to

        Returns:
            True iff the number and names of the locus components, and the component connectivity of the DQA
            match the static architecture of this backend.
        """
```

src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -16,81 +16,112 @@
...
+    * Each gate type is given a single mock implementation.
+    * Calibration set ID is set to the all-zeros UUID.
+
+    Args:
+        sqa: static quantum architecture to replicate
+    Returns:
+        DQA replicating the properties of ``sqa``
+    """
+    qubits = [qb for qb in sqa.qubits if qb.startswith('QB')]
+    computational_resonators = [qb for qb in sqa.qubits if qb.startswith('COMP')]
```
smite, line 53:
```suggestion
    # NOTE this prefix-based heuristic for identifying the qubits and resonators is not guaranteed to work
    qubits = [qb for qb in sqa.qubits if qb.startswith('QB')]
    computational_resonators = [qb for qb in sqa.qubits if qb.startswith('COMP')]
```

src/iqm/qiskit_iqm/iqm_backend.py
```diff
@@ -16,81 +16,112 @@
...
-        qb_to_idx = {qb: idx for idx, qb in enumerate(sorted(architecture.qubits, key=get_num_or_zero))}
-        operations = architecture.operations
-
+        if isinstance(architecture, QuantumArchitectureSpecification):
+            arch = _dqa_from_static_architecture(architecture)
+        else:
+            arch = architecture
+        self.architecture = arch
+
+        qb_to_idx = {qb: idx for idx, qb in enumerate(arch.components)}
```
smite, line 91:
```suggestion
        # Qiskit uses integer indices to refer to qubits
        qb_to_idx = {qb: idx for idx, qb in enumerate(arch.components)}
```

src/iqm/qiskit_iqm/iqm_job.py
```diff
@@ -167,7 +170,7 @@ def result(self) -> Result:
...
                 self.circuit_metadata = []
                 self.circuit_metadata = [c.metadata for c in results.metadata.request.circuits]
```
smite, line 175:
```suggestion
            if self.circuit_metadata is None and results.metadata.request is not None:
                self.circuit_metadata = [c.metadata for c in results.metadata.request.circuits]
```

src/iqm/qiskit_iqm/iqm_job.py
```diff
@@ -167,7 +170,7 @@ def result(self) -> Result:
             self._result = self._format_iqm_results(results)
             # RemoteIQMBackend.run() populates circuit_metadata, so it may be None if method wasn't called in current
             # session. In that case retrieve circuit metadata from RunResult.metadata.request.circuits[n].metadata
```
smite, line 172:
```suggestion
            # IQMBackend.run() populates IQMJob.circuit_metadata, so it may be None if this IQMJob was created manually from a job_id.
            # In that case retrieve circuit metadata from RunResult.metadata.request.circuits[n].metadata
```
I guess this is hat the comment is trying to say?

src/iqm/qiskit_iqm/iqm_provider.py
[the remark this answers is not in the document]
smite: Right, a backend cannot easily automatically track the default calibration on the server, since the transpilation target depends on the calset, and transpilation is a separate operation from run. I guess `transpile` could use a callback to ask the backend to update to the current default calset and Target, but then it could again be out of date when the user calls `run`. I think it's fine like this for now.

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -176,13 +183,29 @@ def create_run_request(self, run_input: Union[QuantumCircuit, list[QuantumCircui
...
+            default_calset_id = self.client.get_dynamic_quantum_architecture(None).calibration_set_id
+            if self._calibration_set_id != default_calset_id:
+                warnings.warn(
+                    f'Server default calibration set has changed from {self._calibration_set_id} '
+                    f'to {default_calset_id}. Use a new IQMBackend to transpile the circuits using '
+                    f'the new calibration set.'
```
smite, line 192:
```suggestion
                    f'to {default_calset_id}. Create a new IQMBackend to transpile the circuits using '
                    f'the new calibration set.'
```

docs/user_guide.rst
smite: Applied

src/iqm/qiskit_iqm/iqm_backend.py
smite: Applied

src/iqm/qiskit_iqm/iqm_job.py
smite: Applied

src/iqm/qiskit_iqm/iqm_provider.py
smite: Applied.

src/iqm/qiskit_iqm/fake_backends/iqm_fake_backend.py
smite: Applied

src/iqm/qiskit_iqm/iqm_backend.py
smite: Applied

PR #125, opened 2024-10-30
Fix the IQMFakeDeneb readout error value

svmjdr:
The ErrorProfile for the IQMFakeDeneb backend had readout fidelity numbers in place of readout error. This PR fixes that problem.


Aerylia:
Can you update the CHANGELOG.rst with a description of your change and a newer version.

For example:
```
Version 13.17
=============

* Fixed Deneb Readout errors to closer resemble reality. Fidelities were reported as errors. `#125 <https://github.com/iqm-finland/iqm-client/pull/125>`_

```

If you want to make more involved changes in the future, you can use `tox -e format` to automatically fix the files to fit our formatting guidelines. After that, you can run `tox` to automatically check if the tests pass locally on your computer. Later, Github will run these tests for other supported OSs before the changes can be merged.

Aerylia (changes requested):
You accidentally commited a bunch of .idea files that should not be in the repository. Can you remove them?

.DS_Store
Aerylia, line 1: This file also needs to be removed, since this is user-specific local data.

Aerylia (changes requested):

Aerylia:
The tests are currently failing, this is because some tests were using private fields in `iqm_client` which they were not supposed to use and now no longer exist. This problem is fixed by another PR that is currently under review. The plan is to wait for those changes, then update the branch with latest, and then rerun the tests. This might take a few days.

Aerylia:
The other PRs are merged. You will have to merge main into your branch. This will create a merge conflict in the CHANGELOG.rst that you will have to resolve manually. The result should be the changelog from main with your change added as a new version, which is now version 15.1

Aerylia:
I've incorporated these changes into #116 so this is obsolete.

PR #126, opened 2024-11-01
Extract instruction serialization

Jajiko:
- Extract most of the instruction conversion logic from a backend method to a pure function. The old behavior is not changed but  GitHub does not show the changes neatly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ Changelog
 Version 15.1
 ============
 
-* Expose part of instruction serialization as :func:`iqm.qiskit_iqm.iqm_provider.serialize_instructions`.
+* Move a part of circuit serialization into :func:`iqm.qiskit_iqm.iqm_provider._serialize_instructions`.
   `#126 <https://github.com/iqm-finland/qiskit-on-iqm/pull/126>`_
 
 Version 15.0
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Version 15.1
+============
+
+* Expose part of instruction serialization as :func:`iqm.qiskit_iqm.iqm_provider.serialize_instructions`.
+  `#126 <https://github.com/iqm-finland/qiskit-on-iqm/pull/126>`_
+
 Version 15.0
 ============
 
--- a/src/iqm/qiskit_iqm/iqm_provider.py
+++ b/src/iqm/qiskit_iqm/iqm_provider.py
@@ -241,109 +241,7 @@ class IQMBackend(IQMBackendBase):
         Raises:
             ValueError: circuit contains an unsupported instruction or is not transpiled in general
         """
-        # pylint: disable=too-many-branches,too-many-statements
-        instructions: list[Instruction] = []
-        # maps clbits to the latest "measure" instruction to store its result there
-        clbit_to_measure: dict[Clbit, Instruction] = {}
-
-        for circuit_instruction in circuit.data:
-            instruction = circuit_instruction.operation
-            qubit_names = [str(circuit.find_bit(qubit).index) for qubit in circuit_instruction.qubits]
-            if instruction.name == 'r':
-                angle_t = float(instruction.params[0] / (2 * np.pi))
-                phase_t = float(instruction.params[1] / (2 * np.pi))
-                native_inst = Instruction(name='prx', qubits=qubit_names, args={'angle_t': angle_t, 'phase_t': phase_t})
-            elif instruction.name == 'x':
-                native_inst = Instruction(name='prx', qubits=qubit_names, args={'angle_t': 0.5, 'phase_t': 0.0})
-            elif instruction.name == 'rx':
-                angle_t = float(instruction.params[0] / (2 * np.pi))
-                native_inst = Instruction(name='prx', qubits=qubit_names, args={'angle_t': angle_t, 'phase_t': 0.0})
-            elif instruction.name == 'y':
-                native_inst = Instruction(name='prx', qubits=qubit_names, args={'angle_t': 0.5, 'phase_t': 0.25})
-            elif instruction.name == 'ry':
-                angle_t = float(instruction.params[0] / (2 * np.pi))
-                native_inst = Instruction(name='prx', qubits=qubit_names, args={'angle_t': angle_t, 'phase_t': 0.25})
-            elif instruction.name == 'cz':
-                native_inst = Instruction(name='cz', qubits=qubit_names, args={})
-            elif instruction.name == 'move':
-                native_inst = Instruction(name='move', qubits=qubit_names, args={})
-            elif instruction.name == 'barrier':
-                native_inst = Instruction(name='barrier', qubits=qubit_names, args={})
-            elif instruction.name == 'measure':
-                if len(circuit_instruction.clbits) != 1:
-                    raise ValueError(
-                        f'Unexpected: measurement instruction {circuit_instruction} uses multiple classical bits.'
-                    )
-                clbit = circuit_instruction.clbits[0]  # always a single-qubit measurement
-                mk = str(MeasurementKey.from_clbit(clbit, circuit))
-                native_inst = Instruction(name='measure', qubits=qubit_names, args={'key': mk})
-                clbit_to_measure[clbit] = native_inst
-            elif instruction.name == 'reset':
-                # implemented using a measure instruction to measure the qubits, and
-                # one cc_prx per qubit to conditionally flip it to |0>
-                feedback_key = '_reset'
-                instructions.append(
-                    Instruction(
-                        name='measure',
-                        qubits=qubit_names,
-                        args={
-                            # HACK to get something unique, remove when key can be omitted
-                            'key': f'_reset_{len(instructions)}',
-                            'feedback_key': feedback_key,
-                        },
-                    )
-                )
-                for q in qubit_names:
-                    physical_qubit_name = self._idx_to_qb[int(q)]
-                    instructions.append(
-                        Instruction(
-                            name='cc_prx',
-                            qubits=[q],
-                            args={
-                                'angle_t': 0.5,
-                                'phase_t': 0.0,
-                                'feedback_key': feedback_key,
-                                'feedback_qubit': physical_qubit_name,
-                            },
-                        )
-                    )
-                continue
-            elif instruction.name == 'id':
-                continue
-            else:
-                raise ValueError(
-                    f"Instruction '{instruction.name}' in the circuit '{circuit.name}' is not natively supported. "
-                    f'You need to transpile the circuit before execution.'
-                )
-            # classically controlled gates (using the c_if method)
-            condition = instruction.condition
-            if condition is not None:
-                if native_inst.name != 'prx':
-                    raise ValueError(
-                        'This backend only supports conditionals on r, x, y, rx and ry gates,'
-                        f' not on {instruction.name}'
-                    )
-                native_inst.name = 'cc_prx'
-                creg, value = condition
-                if len(creg) != 1:
-                    raise ValueError(f'{instruction} is conditioned on multiple bits, this is not supported.')
-                if value != 1:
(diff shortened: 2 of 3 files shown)


src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -241,119 +241,148 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
+            )
+            metadata = None
+
+        return Circuit(name=circuit.name, instructions=instructions, metadata=metadata)
+
+
+def serialize_instructions(
+    circuit: QuantumCircuit, index_to_qubit_mapping: dict[int, str], ignore_nonnative_gates: bool
+) -> list[Instruction]:
+    """Serialize a quantum circuit into Instructions in the IQM data transfer format.
```
smite, line 260:
```suggestion
    """Serialize a quantum circuit into the IQM data transfer format.
```

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -241,119 +241,148 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
+
+    This is an internal helper for :meth:`.IQMBackend.serialize_circuit` that gives slightly more control.
+    See :meth:`.IQMBackend.serialize_circuit` for details.
+    There is usually no need to use this function directly.
+
+    Args:
+        circuit: quantum circuit to serialize
+        index_to_qubit_mapping: Mapping from logical to physical qubit names.
```
smite, line 268:
```suggestion
        index_to_qubit_mapping: Mapping from qubit indices to the corresponding qubit names.
```
I think there are multiple use cases, some of which use logical and some of which use physical indices/names. Better to be imprecise.

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -241,119 +241,148 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
+        except ValueError:
+            warnings.warn(
+                f'Metadata of circuit {circuit.name} was dropped because it could not be serialised to JSON.',
+            )
+            metadata = None
+
+        return Circuit(name=circuit.name, instructions=instructions, metadata=metadata)
+
+
+def serialize_instructions(
```
freetonik, line 257: Now that Pulse-level access library can use this pure function, it will still have to hackishly convert it into another (though equivalent) class. Not a huge deal I guess, but ultimately if the core Pulse library were open-source, IQM Client could re-use that data structure instead of defining it own `Instruction`.

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -241,119 +241,148 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
+    There is usually no need to use this function directly.
+
+    Args:
+        circuit: quantum circuit to serialize
+        index_to_qubit_mapping: Mapping from logical to physical qubit names.
+        ignore_nonnative_gates: If False (default), any instruction that cannot be serailizaed into the native
+            gate set will raise an error. If True, such instructions are converted as-is without validation,
+            and the caller must edit the result to be valid and executable.
+            Notably, since IQM transfer format requires named parameters and qiskit parameters don't have names, the
+            `i` th parameter of an unrecognized instruction is given the name ``"param<i>"``.
```
smite, line 273: Would it be better if this was a list of allowed gate names? Then we could retain some control over transpilation errors etc.

smite: I mean allowed on top of the hardcoded ones in the if/elif clauses.

src/iqm/qiskit_iqm/iqm_provider.py
Jajiko: yes, agree

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -241,119 +241,148 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
+
+        try:
+            metadata = to_json_dict(circuit.metadata)
+        except ValueError:
+            warnings.warn(
+                f'Metadata of circuit {circuit.name} was dropped because it could not be serialised to JSON.',
+            )
+            metadata = None
+
+        return Circuit(name=circuit.name, instructions=instructions, metadata=metadata)
```
smite, line 254: Can we make it `serialize_circuit`, so the pure func would return a `Circuit` object (with the correct name), and after the try block it would just add metadata into the existing object?

src/iqm/qiskit_iqm/iqm_provider.py
Jajiko: I feel it's unnecessary complication for the user.
A Pulla user already has to be explicit about registering new gates. If a gate is not registered, the will get the "Gate not registered" error from iqm-pulse.

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -241,119 +241,148 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
-                    )
-                continue
-            elif instruction.name == 'id':
-                continue
+            continue
+        elif instruction.name == 'id':
+            continue
+        else:
+            if ignore_nonnative_gates:
+                args = {f"param{i}": param for i, param in enumerate(instruction.params)}
```
smite, line 351:
```suggestion
                args = {f"p{i}": param for i, param in enumerate(instruction.params)}
```

smite: Also pylint will probably suggest making that "else: if:" into another "elif:"

src/iqm/qiskit_iqm/iqm_provider.py
Jajiko: ok

src/iqm/qiskit_iqm/iqm_provider.py
smite: Yes, but some of those gates may come from Qiskit itself if they forgot to transpile the circuit first. It may not be clear to the user why they were not accepted by iqm-pulse. Whereas if we use a user-submitted whitelist, those gates will be passed but the serialization function will give a helpful error message about the others.

src/iqm/qiskit_iqm/iqm_provider.py
smite: So it will try to revalidate the list of Instructions which already have been validated, when you construct the Circuit? How does it work now?

src/iqm/qiskit_iqm/iqm_provider.py
Jajiko: Changed.

src/iqm/qiskit_iqm/iqm_provider.py
Jajiko: Yes it validates all input Instructions recursively. I suppose it has no way of knowing they have already been validated before.

Jajiko: In my current version, we never create a Circuit with invalid Instructions.

src/iqm/qiskit_iqm/iqm_provider.py
Jajiko: Changed.

CHANGELOG.rst
```diff
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Version 15.1
+============
+
+* Expose part of instruction serialization as :func:`iqm.qiskit_iqm.iqm_provider.serialize_instructions`.
```
smite, line 8:
```suggestion
* Move a part of circuit serialization into :func:`iqm.qiskit_iqm.iqm_provider._serialize_instructions`.
```

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -241,119 +241,147 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
+    circuit: QuantumCircuit, index_to_qubit_mapping: dict[int, str], allowed_nonnative_gates: Collection[str] = ()
+) -> list[Instruction]:
+    """Serialize a quantum circuit into the IQM data transfer format.
+
+    This is IQM's internal helper for :meth:`.IQMBackend.serialize_circuit` that gives slightly more control.
+    See :meth:`.IQMBackend.serialize_circuit` for details.
+
+    Args:
+        circuit: quantum circuit to serialize
+        index_to_qubit_mapping: Mapping from qubit indices to the corresponding qubit names.
```
smite, line 267:
```suggestion
    circuit: QuantumCircuit,
    qubit_index_to_name: dict[int, str],
    allowed_nonnative_gates: Collection[str] = ()
) -> list[Instruction]:
    """Serialize a quantum circuit into the IQM data transfer format.

    This is IQM's internal helper for :meth:`.IQMBackend.serialize_circuit` that gives slightly more control.
    See :meth:`.IQMBackend.serialize_circuit` for details.

    Args:
        circuit: quantum circuit to serialize
        qubit_index_to_name: Mapping from qubit indices to the corresponding qubit names.
```

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -241,119 +241,147 @@ def serialize_circuit(self, circuit: QuantumCircuit) -> Circuit:
...
+            continue
+        else:
+            if instruction.name in allowed_nonnative_gates:
+                args = {f'p{i}': param for i, param in enumerate(instruction.params)}
+                native_inst = Instruction.model_construct(name=instruction.name, qubits=qubit_names, args=args)
             else:
                 raise ValueError(
                     f"Instruction '{instruction.name}' in the circuit '{circuit.name}' is not natively supported. "
                     f'You need to transpile the circuit before execution.'
                 )
```
smite, line 356:
```suggestion
        elif instruction.name in allowed_nonnative_gates:
            args = {f'p{i}': param for i, param in enumerate(instruction.params)}
            native_inst = Instruction.model_construct(name=instruction.name, qubits=qubit_names, args=args)
        else:
            raise ValueError(
                f"Instruction '{instruction.name}' in the circuit '{circuit.name}' is not natively supported. "
                f'You need to transpile the circuit before execution.'
            )
```

tests/test_iqm_provider.py
```diff
@@ -119,3 +119,19 @@ def test_facade_backend_raises_error_on_remote_execution_fail(adonis_architectur
...
+    circuit.append(nonnative_gate, [1, 2, 4])
+    circuit.measure_all()
+
+    with pytest.raises(ValueError, match='is not natively supported. You need to transpile'):
+        _serialize_instructions(circuit, {i: f'QB{i+1}' for i in range(5)})
+
+    instructions = _serialize_instructions(
+        circuit, {i: f'QB{i+1}' for i in range(5)}, allowed_nonnative_gates={'nonnative'}
+    )
+    assert instructions[0] == Instruction.model_construct(name='nonnative', qubits=['1', '2', '4'], args={})
```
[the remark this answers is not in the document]
Jajiko: All other instructions use `qubit_names = [str(circuit.find_bit(qubit).index) for qubit in circuit_instruction.qubits]`, and this follows the same pattern. The qubit mapping in the signature is used only for `cc_prx` and `measure`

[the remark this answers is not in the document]
Jajiko: TBH I'm quite lost with the meaning of these mappings.

src/iqm/qiskit_iqm/iqm_provider.py
Jajiko: Applied.

src/iqm/qiskit_iqm/iqm_provider.py
Jajiko: Applied.

tests/test_iqm_provider.py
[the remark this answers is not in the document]
smite: We'll try to eliminate some soon.

smite (approved):

PR #131, opened 2024-11-14
Fix list in user guide

freetonik:

PR #132, opened 2024-11-20
Do not use original request directly

kukushechkin:
- This fixes compatibility with iqm-client APIVariant V2.
- support iqm-client 20.5 in unit tests
- make resonance example testable, an improvement over @manzanillo implementation from https://github.com/iqm-finland/qiskit-on-iqm/pull/131
- cherry-pick resonance guide by @freetonik from https://github.com/iqm-finland/qiskit-on-iqm/pull/131

I suggest also closing https://github.com/iqm-finland/qiskit-on-iqm/pull/131 if this one is merged.

### DIFF
--- a/src/iqm/qiskit_iqm/examples/resonance_example.py
+++ b/src/iqm/qiskit_iqm/examples/resonance_example.py
@@ -17,31 +17,64 @@ See the Qiskit on IQM user guide for instructions:
 https://iqm-finland.github.io/qiskit-on-iqm/user_guide.html
 """
 
-from iqm.qiskit_iqm import IQMProvider, transpile_to_IQM
+import argparse
+from typing import Optional
+
 from qiskit import QuantumCircuit
 
-server_url = "https://cocos.resonance.meetiqm.com/<QUANTUM COMPUTER>"  # For example https://cocos.resonance.meetiqm.com/garnet
-api_token = "<INSERT YOUR TOKEN>"
+from iqm.qiskit_iqm import IQMProvider, transpile_to_IQM
 
-SHOTS = 1000
 
-# Define a quantum circuit
-num_qb = 5
-qc = QuantumCircuit(num_qb)
+def resonance_example(server_url: str, api_token: Optional[str]) -> dict[str, int]:
+    """Run a circuit via IQM Resonance.
 
-qc.h(0)
-for qb in range(1, num_qb):
-    qc.cx(0, qb)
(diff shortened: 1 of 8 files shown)


tests/test_iqm_provider.py
```diff
@@ -113,6 +125,9 @@ def test_facade_backend_raises_error_on_remote_execution_fail(adonis_architectur
     when(IQMClient).submit_run_request(...).thenReturn(uuid.uuid4())
     when(IQMClient).get_run(ANY(uuid.UUID)).thenReturn(RunResult.from_dict(result))
     when(IQMClient).get_run_status(ANY(uuid.UUID)).thenReturn(RunStatus.from_dict(result_status))
+    when(requests).get('http://some_url/info/client-libraries', headers=matchers.ANY, timeout=matchers.ANY).thenReturn(
+        get_mock_ok_response({'qiskit-iqm': '1.2.3'})
```
antonsergeyev, line 129: a valid response that iqm-client expects is
```
{
  'iqm-client': {
      'min': '10.0',
      'max':'30.0',
  }
}
```

your response works too because iqm-client ignores parsing errors, but it's kind of misleading

src/iqm/qiskit_iqm/examples/resonance_example.py
```diff
@@ -17,31 +17,65 @@
...
+        default='<INSERT YOUR TOKEN>',
+    )
 
-# Run the circuit
-job = backend.run(qc_transpiled, shots=SHOTS)
-print(job.result().get_counts())
+    args = argparser.parse_args()
+    circuit_transpiled, results = resonance_example(args.url, args.token)
+    print(circuit_transpiled)
+    print(results)
```
jkotilahti, line 81: `resonance_example()` doesn't return transpiled circuit, only results, so I think this should be
```suggestion
    results = resonance_example(args.url, args.token)
    print(results)
```

jkotilahti (changes requested):

src/iqm/qiskit_iqm/examples/resonance_example.py
jkotilahti: (or alternatively make `resonance_example` return the transpiled circuit)

kukushechkin: Oh yeah, good catch, updated.

jkotilahti (approved):

PR #140, opened 2024-12-12
CZ implementations may be nonsymmetric, only calibrated locus orders must be used.

Aerylia:


Aerylia:
I kept the symmetry argument when creating the Target in case we get (other) symmetric gates again.

CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 15.6
+============
+
+* Qiskit Target now contains CZ with the directions are calibrated on the hardwared. `#140 <https://github.com/iqm-finland/qiskit-on-iqm/pull/140>`_
```
smite, line 8:
```suggestion
* CZ gates now appear in the transpilation target only in the direction they are calibrated. `#140 <https://github.com/iqm-finland/qiskit-on-iqm/pull/140>`_
```

Aerylia:
TBH, I don't think this is relevant. The compiler can handle this case and it will be overwritten by #116

Aerylia:
This was part of #116

PR #141, opened 2024-12-12
Add lock file

shabeebk:
Vulnerability scanning tooling is using the lock file to look for issues. Without this, the tooling will not be able to produce reliable report.


.pre-commit-config.yaml
Aerylia, line 1: Could you add tox -e format here too? Then pre-commit is setup completely immediately

shabeebk: Done.

shabeebk:
> You’ll need to update the changelog. Minor release.
> 
> * Added a requirements lock file that is automatically generated with tox and pre-commit

Added.

CHANGELOG.rst
```diff
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+Version 15.6
+============
+
+* Added requirements lock file that is automatically generated with tox and pre-commit.
```
miikkako, line 8: requirements.txt seems not used anywhere?

miikkako: Or how is the vulnerability scanner configured to use the lockfile? Either way if it's only used by CI and not even included in the published python package I don't think a changelog is needed

Aerylia: I asked for one to track these changes in case some external dev runs into them. As we often have externals making PRs on this repo.

shabeebk: The requirements.txt is used by the Aikido app integration and hence will not be visible here. And, the changelog is added as @Aerylia asked for it.

miikkako: is requirements.txt part of the published python package?

miikkako: Well perhaps it doesn't matter but I'd consider leaving it out to avoid confusion

Aerylia: Ok, it can be removed.

shabeebk: Reverted. We can squash the commits while merging to make it clean.

miikkako: I don't see any benefit for having this lockfile if the python package and CI are not using it. I think we should either make the package depend on the lockfile, although for client libraries this wouldn't make much sense and would actually be a big nuisance for runtimes depending on this library. The other option is to simply ignore this aikido complaint which I think is correct for client library repos.

.github/workflows/ci.yml
```diff
@@ -29,14 +29,32 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -e ".[dev]"
+          python -m pip install tox-gh-actions==2.12.0
+      - name: Regenerate requirements.txt
+        run: tox -e pip-compile
+      - name: Check if requirements.txt is up-to-date
+        run: git diff --exit-code HEAD requirements.txt
+
```
shabeebk, line 49:
```suggestion
```

shabeebk (commented):
Remove the lock file check in the CI.

Aerylia (approved):

jkotilahti:
Agreed with @shabeebk that this can be closed.

PR #145, opened 2025-02-06
Reflect Resonance authentication in docs better

manzanillo:
This pull request updates the authentication process in `IQMProvider` and improves the user guide to better reflect the usage of IQM Resonance.

Changes:
* Updated IQMProvider to mention authentication option for IQM Resonance 
* Mentioned the support for the token parameter in the User guide.
* Added separating headlines for IQM Resonance and on-prem QCs

### DIFF
--- a/src/iqm/qiskit_iqm/iqm_provider.py
+++ b/src/iqm/qiskit_iqm/iqm_provider.py
@@ -433,20 +433,13 @@ class IQMFacadeBackend(IQMBackend):
 class IQMProvider:
     """Provider for IQM backends.
 
-    Args:
-        url: URL of the IQM Quantum Computer (e.g. https://cocos.resonance.meetiqm.com/garnet)
+    IQMProvider connects to a quantum computer through an IQM server.
+    If the server requires user authentication, you can provide it either using environment
+    variables, or as keyword arguments to IQMProvider. The user authentication kwargs are passed
+    through to :class:`~iqm.iqm_client.iqm_client.IQMClient` as is, and are documented there.
 
-    Keyword Args:
-        auth_server_url: URL of the user authentication server, if required by the IQM Cortex server, not required for IQM Resonance.
-            Can also be set in the ``IQM_AUTH_SERVER`` environment variable.
-        username: Username, if required by the IQM Cortex server, not required for IQM Resonance.
-            Can also be set in the ``IQM_AUTH_USERNAME`` environment variable.
-        password: Password, if required by the IQM Cortex server, not required for IQM Resonance.
-            Can also be set in the ``IQM_AUTH_PASSWORD`` environment variable.
-        password: Password, if required by the IQM Cortex server, not required for IQM Resonance.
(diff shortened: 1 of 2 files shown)


src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -434,15 +434,19 @@ class IQMProvider:
     """Provider for IQM backends.
 
     Args:
-        url: URL of the IQM Cortex server
+        url: URL of the IQM Quantum Computer (e.g. https://cocos.resonance.meetiqm.com/garnet)
 
     Keyword Args:
         auth_server_url: URL of the user authentication server, if required by the IQM Cortex server.
```
manzanillo, line 440:
```suggestion
        auth_server_url: URL of the user authentication server, if required by the IQM Cortex server, not required for IQM Resonance.
```

src/iqm/qiskit_iqm/iqm_provider.py
```diff
@@ -433,16 +433,13 @@ def run(self, run_input: Union[QuantumCircuit, list[QuantumCircuit]], **options)
...
     Args:
-        url: URL of the IQM Cortex server
-
-    Keyword Args:
-        auth_server_url: URL of the user authentication server, if required by the IQM Cortex server.
-            Can also be set in the ``IQM_AUTH_SERVER`` environment variable.
-        username: Username, if required by the IQM Cortex server.
-            Can also be set in the ``IQM_AUTH_USERNAME`` environment variable.
-        password: Password, if required by the IQM Cortex server.
-            Can also be set in the ``IQM_AUTH_PASSWORD`` environment variable.
```
freetonik, line 445: 1. these docstrings are removed with the assumption that the user would, if needed, discover them in `IQMClient`? 
2. overall, the updated docstring is applicable to Resonance only; on-premise users would find this confusing, I think.

smite: There's  direct link to IQMClient docs, where we explain what all the auth options do. The idea is to keep the docs in one place, where they are easier to keep up to date.

How is the new docstring applicable to resonance only?

freetonik: >How is the new docstring applicable to resonance only?

It removes username/password from the docstring. Without visiting the IQM Client documentation, a user might not realize username+password option exists.

smite: I removed *all* details on the auth parameters from the docstring, now it just says they exist, and are documented in IQMClient, and there's a direct intersphinx link to the IQMClient documentation. How is it specific to Resonance?

manzanillo: If we remove the details from the docstring to make maintenance easier we should mention Resonance/on-prem differences in the IQM Client documentation though. But I understand the point of avoiding the duplication

freetonik: For posterity: discussed and resolved AFK.

PR #147, opened 2025-02-20
SW-803: Bugfix in IQMOptimizeSingleQubitGates where the angles are not properly computed 

Aerylia:
Bugfix in ``IQMOptimizeSingleQubitGates`` where the angles are not properly computed for circuits with prx gates on qubits holding a resonator state, i.e. circuits that require running without move gate validation.


perjlieb (approved):
LGTM

PR #149, opened 2025-02-21
add warning when unknown facade name is used

manzanillo:
This PR introduces a warning in case a facade backend name is used that does not exist, I did not add a release note, it can be released with whatever else comes next


JMuff22:
Thanks @manzanillo !

freetonik:
Do you want to make a release? If so, please add a changelog entry.

freetonik (approved):

PR #151, opened 2025-02-27
Bugfix: transpile() fails in a circuit with delay ops.

smite:
* `IQMOptimizeSingleQubitGates` now accepts `delay` operations.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Version 17.4
+============
+
+* Bugfix: The ``delay`` operation is now accepted by the standard transpiler.
+  `#151 <https://github.com/iqm-finland/qiskit-on-iqm/pull/151>`_
+
 Version 17.3
 ============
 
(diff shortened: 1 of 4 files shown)


Aerylia:
Hmm. Is there a nice way to make this fix more general rather than hardcode which gates commute with Rzs? So that with the next new gate, the fix still works

smite:
> Hmm. Is there a nice way to make this fix more general rather than hardcode which gates commute with Rzs? So that with the next new gate, the fix still works

Ideally the entire single-qubit gate optimization pass should be included in Qiskit itself.
The Qiskit transpiler should know which circuit operations commute with one another, since it has [this pass](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.passes.CommutationAnalysis), maybe we could use it instead of mapping gate names explicitly to their RZ commutation properties in the if/elif.

Aerylia:
Yeah, something along those lines. Let's make it a separate ticket, so that this can be merged.

Aerylia (approved):
lgtm.
Needs an additional ticket for automatic checking of the gate commutation in the optimization pass.

PR #156, opened 2025-03-09
Update Apollo's noise parameters to resemble Garnet's specs

jungthapa:
This PR updates the noise parameters for Apollo to better resemble the specifications of Garnet. This ensures that simulations using Apollo's noise model are more accurate and reflective of Garnet.

Changes:

Updated noise parameters in fake_apollo.py to match Garnet's specifications -- calibration data used from 5.3.2025.


jkotilahti (changes requested):
The changes look good, but please add also a changelog entry for this.

jkotilahti (changes requested):

jungthapa:
@jkotilahti Thanks for your review. One generic question: Are we supposed to make fake_garnet.py script or is update on fake_apollo.py fine? Because Apollo is a family of QPU and Garnet is a specific one from that family. What should the standard be going forward?

jungthapa:
Ok. I would then actually opt for adding IQMFakeGarnet. Let me update the PR asap.

jungthapa:
Oops forgot to update changelog

src/iqm/qiskit_iqm/fake_backends/fake_apollo.py
```diff
@@ -215,4 +215,4 @@ def IQMFakeApollo() -> IQMFakeBackend:
         name="sample-chip",
     )
 
-    return IQMFakeBackend(architecture, error_profile, name="IQMFakeApolloBackend")
+    return IQMFakeBackend(architecture, error_profile, name="IQMFakeApolloBackend")
```
jkotilahti, line 218: The last (empty) line was accidentally removed, it needs to be added back for the formatting checks to pass.
```suggestion
    return IQMFakeBackend(architecture, error_profile, name="IQMFakeApolloBackend")

```

src/iqm/qiskit_iqm/fake_backends/fake_garnet.py
```diff
@@ -0,0 +1,218 @@
...
+        operations={
+            "prx": [[q] for q in qubits],
+            "cz": list(qubit_connectivity),
+            "measure": [[q] for q in qubits],
+            "barrier": [],
+        },
+        qubits=qubits,
+        qubit_connectivity=qubit_connectivity,
+    )
+    # Note that these specs are ballpark numbers and don't correspond directly to a specific device
```
jkotilahti, line 87: This comment was valid for FakeApollo, but should be removed from FakeGarnet.
```suggestion
```

CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 17.6
+============
+
+* Added Garnet backend so that it can be used in simulations.
```
jkotilahti, line 8: Not very important, but good to add also link to the pull request, similar to other changelog entries.
```suggestion
* Added Garnet backend so that it can be used in simulations.
  `#156 <https://github.com/iqm-finland/qiskit-on-iqm/pull/156>`_
```

jkotilahti (changes requested):
Please add also `test_fake_garnet.py`. You can just copy `test_fake_apollo.py` and replace "apollo" with "garnet" everywhere.

tests/fake_backends/test_fake_garnet.py
```diff
@@ -0,0 +1,71 @@
...
+from iqm.qiskit_iqm.iqm_backend import IQMTarget
+
+
+def test_iqm_fake_garnet():
+    backend = IQMFakeGarnet()
+    assert backend.num_qubits == 20
+    assert backend.name == 'IQMFakeGarnetBackend'
+
+
+def test_iqm_fake_garnet_connectivity():
```
jkotilahti, line 29: Looks like `pylint` didn't like the copy-pasted code. In the future we want to change things so that this kind of copy-paste is not needed for fake backends, but for now you can just disable the pylint check:
```suggestion
def test_iqm_fake_garnet_connectivity():
    # pylint: disable=duplicate-code
```
This should make the CI pipeline pass, and then I will approve.

jungthapa:
Oops. something failed in CI. Let me check.

jungthapa:
hmm CI is not passing because of similar lines. But this change is something you suggested and I applied it. I am a bit confused.

jkotilahti:
Sorry, it looks like my local testing didn't detect all the places where that failure happens. I think you need to add the `# pylint: disable=duplicate-code` also at least to IQMFakeGarnet. 

You can run `tox --recreate` locally to test that the checks pass. (I ran just `tox` initially, but it falsely passed due to caching, using the `--recreate` option seems necessary to properly check that everything passes.)

jkotilahti (approved):

PR #158, opened 2025-03-17
Symbolic gates transpilation

Aerylia:
Added the hack to support symbolic gate transpilation.

Note that tests for python 3.10 are failing. This is because of a bug in iqm-client that is fixed in version 22.7, but does not work because of support was dropped in 22.5


src/iqm/qiskit_iqm/iqm_naive_move_pass.py
```diff
@@ -66,6 +68,19 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
             TranspilerError: The layout is not compatible with the DAG, or if the input gate set is incorrect.
         """
+        # pylint: disable=too-many-branches
+
+        # Replace symbolic parameters with indices and store the index to symbol mapping.
+        symbolic_gates = {}
+        symbolic_index = 0
+        for node in dag.topological_op_nodes():
+            # This only works for prx gates because that has two parameters
+            # We use one to mark that it is a symbolic gate (np.inf) and the other to store the index.
```
SalahedeenIssa, line 78: Is there a way to tag this using an Instruction [label](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.Instruction#label) perhaps, rather than setting the first parameter to infinity to indicate that it's a symbolic gate?

Aerylia: No, then the serialization method breaks because we need to turn the DAG into an `iqm-client.Circuit` object where instructions have very precise typing and prx gates can only have float parameters. We decided not to change the Circuit definition here as it is used by many libaries and should be compatible with our devices that cannot run circuits with unknown parameters in it. This hack is the cleanest solution.

smite: Would it be possible to avoid the Pydantic validation somehow for this use case here? This would require changes in iqm-client though. Maybe it could also be done later.

Aerylia: In principle there are many other hacks, they will just become more complicated. There is a ticket for handling symbolic gates in our circuit definition properly, but iqm-benchmarks is currently blocked.

src/iqm/qiskit_iqm/iqm_naive_move_pass.py
```diff
@@ -66,6 +68,19 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            if node.name == "r" and not all(isinstance(param, (float, int)) for param in node.op.params):
```
SalahedeenIssa, line 79: Forgive me if this is a silly question, but what would the param be if not a float or int? What if a user submits a symbolic gate with only floats or ints as params?

SalahedeenIssa: Got it. Then this LGTM. Could be beneficial to seek additional approval from someone more familiar with this repo though

SalahedeenIssa (approved):

src/iqm/qiskit_iqm/iqm_naive_move_pass.py
freetonik, line 78: Maybe worth adding a `TODO` or similar comment about the temporary nature of `inf`, and the fact that we should introduce an intermediate circuit representation to properly handle symbolic parameters.

tests/test_iqm_naive_move_pass.py
```diff
@@ -401,3 +401,33 @@ def test_transpile_to_IQM_flags(
...
+    p_vec = ParameterVector("p", length=2)
+
+    qc = QuantumCircuit(2)
+    qc.rz(p_vec[0], 0)
+    qc.rz(p_vec[1], 1)
+    qc.cz(0, 1)
+    qc.measure_all()
+
+    # Transpiling should not error out
+    qc_t1 = transpile_method(qc, backend=backend)
```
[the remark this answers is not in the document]
Aerylia: First, this case also pops up with Qiskit's `transpile` method, as they both call our transpiler passes by default.

For transpilation, it should not matter which of the parameters has been assigned. If you try to run the circuit with unassigned parameters, the serialization will fail with this "Unbound parameter" error. We want this for now as we need to know what the arguments of the gates are when running the circuit. This behavior should be the same as long as there is any unbound parameter in the circuit, so the case you describe is the same as sending a circuit with 1 parameter in the circuit, rather than 2, and thus should be covered by the test.

src/iqm/qiskit_iqm/iqm_naive_move_pass.py
Aerylia: Done

freetonik (approved):

